PR #84, opened 2026-02-03
refactor(ci): self-contained test tasks - each package uses its own vitest

schickling:
## Summary

- Each package now uses its own vitest from `node_modules/.bin/vitest` (self-contained)
- Added `vitest.config.ts` to packages that were missing one
- Removed shared vitest dependency that violated self-contained packages requirements

## Problem

Previously, test tasks were configured to use a shared vitest binary from `packages/@overeng/utils/node_modules/.bin/vitest`. This violated the self-contained packages requirements (R1-R5 in `context/repo-composition/self-contained-packages/requirements.md`):

- R1: Each package must define its own dependencies
- R5: Dependency resolution must be reproducible from the package's lockfile alone

The test tasks had a hidden dependency on the `utils` package being installed first, causing CI failures when the cache was fresh.

## Solution

Adopted a self-contained approach where each package:
1. Has vitest as a devDependency (already the case)
2. Has its own `vitest.config.ts` (added where missing)
3. Runs tests using its own `node_modules/.bin/vitest`

This ensures packages are independently testable without cross-package dependencies.

## Changes

- `nix/devenv-modules/tasks/shared/test.nix`: Simplified to use `node_modules/.bin/vitest run` from each package's directory
- `devenv.nix`: Removed `vitestBin`, `vitestConfig`, and `vitestInstallTask` parameters
- Added `vitest.config.ts` to: effect-path, effect-rpc-tanstack, genie, notion-cli, notion-effect-client, notion-effect-schema

## Note on nix-check failure

The `nix-check` job failure is a pre-existing issue affecting main branch. It's unrelated to this PR and needs separate investigation (likely a cachix cache issue).

### DIFF
--- a/nix/devenv-modules/tasks/shared/test.nix
+++ b/nix/devenv-modules/tasks/shared/test.nix
@@ -25,57 +25,39 @@
 #
 # Provides:
 #   - test:run - Run all tests
-#   - test:watch - Run tests in watch mode (requires single package or vitestBin override)
+#   - test:watch - Run tests in watch mode
 #   - test:<name> - Run tests for specific package (when packages provided)
 {
   packages ? [],
   installTaskPrefix ? "pnpm",
   extraTests ? [],
-  # Legacy: shared vitest binary (deprecated, violates self-containment)
-  # Use only for migration; each package should have its own vitest
-  vitestBin ? null,
-  vitestConfig ? null,
 }:
 { lib, ... }:
 let
   hasPackages = packages != [];
-  hasSharedVitest = vitestBin != null;
-
-  # Build vitest command for simple (non-package) mode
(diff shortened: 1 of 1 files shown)


schickling:
## CI Status

✅ **Fixed by this PR:**
- `test`: All tests pass now (previously failing with "vitest: No such file or directory")
- `lint`: Passes
- `typecheck`: Passes

⚠️ **Pre-existing issue (not related to this PR):**
- `nix-check`: Failing with "deps hash is stale" - this is a pre-existing issue affecting main branch as well. The nix build works locally but fails in CI, likely due to a cachix cache issue. This needs separate investigation.

The nix-check failure affects all CI runs on main and has been ongoing. This PR can be merged to fix the test issue while the nix-check issue is investigated separately.

PR #267, opened 2026-02-22
Sync nested megarepo.lock during lock sync

schickling:
This PR extends mr sync lock-sync to update nested member megarepo.lock files in addition to flake.lock and devenv.lock so nested megarepos do not drift from the parent lock. Nested lock matching now prefers same-name entries with URL verification and falls back to URL matching, while keeping pinned nested entries unchanged and preserving warn-and-continue parse error behavior. It also updates lock-sync output/schema to report megarepo.lock updates, adds integration tests for nested drift, pinned skip, and exclude behavior, and updates the megarepo lock-sync docs plus CHANGELOG. Validation was run with devenv task gates for ts:check, lint:check:oxlint, lint:check:format, and test:megarepo.


schickling:
Closing in favor of #269, which tracks the revised direction: only run nested `megarepo.lock` sync with `mr sync --all`.

_Posted by Codex on behalf of @schickling._

PR #334, opened 2026-03-06
Encode strict internal dependency alignment in genie

### LINKED ISSUE #329: Define strict internal dependency alignment for source-typed workspace families
We should define and encode a clearer dependency policy for internal workspace packages that are consumed as source and rely on nominal type identity.

Problem:
- some shared dependency families are effectively part of the type-level ABI between workspace packages
- when adjacent workspace packages use mixed spec styles or drifted install-time versions for those families, cross-package type identity can break in hard-to-debug ways
- consumer-facing `peerDependencies` still need ranges, but our internal install-time shape needs a stronger consistency rule

What we should do:
- identify the dependency families that must stay strictly aligned across workspace boundaries
- encode that alignment in genie for install-time dependencies
- keep `peerDependencies` ranged for consumers where that remains appropriate
- document why these families are special so the policy is maintainable

Why:
- this reduces nominal type splits and duplicate-package identity problems across workspace boundaries
- it keeps internal builds/typechecks more deterministic
- it preserves consumer compatibility by not unnecessarily narrowing peer ranges

Concrete validated example:
- in `@overeng/notion-cli`, aligning the install-time versions of the `@effect-atom` / Effect / React cluster to the exact versions already pinned by adjacent workspace packages removed a real cross-package type-identity failure while leaving peer ranges intact

Initial scope:
- `@effect-atom/atom`
- `@effect-atom/atom-react`
- `effect`
- `@effect/platform`
- `@effect/platform-node`
- `@effect/rpc`
- `react`
- `react-dom`
- `react-reconciler`
- related type/runtime companions where needed

Related context:
- upstream package-manager issues made this visible during investigation, but the effect-utils issue is broader: define the right internal alignment policy and encode it in generation

---
<sub>Filed by an AI assistant on behalf of @schickling</sub>

schickling-assistant:
## Why

Some workspace packages consume adjacent packages as source while also depending on parts of their peer surface at install time. When those local install-time dependency specs drift from the exact versions already pinned by the upstream workspace package, cross-package type identity becomes less deterministic.

This change encodes a stricter internal alignment rule in genie while keeping consumer-facing `peerDependencies` ranged.

Closes #329.

## What

- add a shared `alignInstallDependencyVersions` helper in `genie/internal.ts`
- use that helper for packages that inherit peer-facing dependency families from adjacent workspace packages:
  - `@overeng/genie`
  - `@overeng/megarepo`
  - `@overeng/notion-cli`
  - `@overeng/effect-schema-form-aria`
  - `@overeng/notion-effect-client`
- regenerate affected `package.json` files
- refresh the affected package-level `pnpm-lock.yaml` files

## How

The helper keeps each package's `devDependencies` explicit, then overrides only the overlapping names that are also exposed as peer dependencies by the referenced workspace package(s). Those overrides come from the upstream package's exact direct dependency pins, not from its peer ranges.

That gives us:

- exact internal install-time alignment for shared type/runtime families
- unchanged ranged `peerDependencies` for consumers
- a single reusable mechanism instead of ad-hoc per-package overrides

## Validation

- `CI=1 dt genie:run --no-tui`
- `dt pnpm:install:genie --no-tui`
- `dt pnpm:install:notion-cli --no-tui`
- `CI=1 dt check:all --no-tui`
- Bun canary with Bun `1.3.10` from `github:NixOS/nixpkgs/master#bun` in a sanitized temp copy:
  - `@overeng/notion-cli`: `bun install --linker isolated` passed, produced a single `@effect-atom/atom@0.4.13+*` store entry, and both `bun x tsc --build` / `bun x tsc --noEmit -p tsconfig.json` passed
  - `@overeng/genie`: `bun install --linker isolated` passed and both `bun x tsc --build` / `bun x tsc --noEmit -p tsconfig.json` passed

---
<sub>Created by an AI assistant on behalf of @schickling</sub>


schickling-assistant:
Closing in favor of #336.

---
<sub>Closed by an AI assistant on behalf of @schickling</sub>

PR #430, opened 2026-03-21
fix(megarepo): skip pre-flight hygiene checks in apply mode

### LINKED ISSUE #423: Store hygiene failure: shared secondary member .git missing during nested mr apply --all
## Bug

When `mr fetch --apply --all` processes nested megarepos that share a secondary member (`overeng-beads-public`), concurrent worktree creation for that member can corrupt the store (missing `.git`). The nested sync then fails the store hygiene check.

This was supposed to be fixed by StoreLock (#415) but may not fully cover worktree creation races.

## Error

```
Store hygiene check failed with 1 error:
  overeng-beads-public: .git not found in worktree at /path/to/store/github.com/overengineeringstudio/overeng-beads-public/refs/heads/main/
  fix: run 'mr apply' to recreate the worktree
```

## When it happens

- A root megarepo has member repos that are themselves megarepos
- Multiple of those nested megarepos share a common secondary member (`overeng-beads-public`)
- `mr fetch --apply --all` syncs everything recursively
- The shared member's worktree gets corrupted during concurrent creation attempts by different nested megarepos

## Repro

See https://github.com/schickling-repros/2026-03-mr-store-hygiene-nested

**Note:** As of `mr` 0.1.0+b245c4e, the repro did not trigger the race in 50 iterations on dev3. The StoreLock fix (#415) may have resolved it, or the race window may require more nested members / slower I/O to reproduce.

## Root cause hypothesis

StoreLock (#415) serializes bare repo clones but worktree creation for the same repo from multiple nested megarepos may still race. When the second nested megarepo tries to create/verify the worktree, it finds a partially-initialized state left by the first.

---

> Filed by Claude on behalf of @schickling

schickling-assistant:
## Summary

Closes #423. Follow-up to #425.

Fixes the deterministic pre-flight hygiene failure when `mr apply --all` processes nested megarepos that share a secondary member. A deeper clone race (#431) remains but is a separate issue.

## Root cause

Pre-flight hygiene checks run before ANY member sync starts, checking ALL members' store state upfront. In `--all` mode, nested megarepo syncs run in parallel (concurrency: 4). When two nested megarepos share a secondary member, one sync's pre-flight check races with the other's worktree creation — observing in-progress or broken state and blocking with `StoreHygieneError`.

## Why removing pre-flight in apply mode is correct

Apply mode self-heals all store issues — `syncMember` clones missing bare repos, recreates broken worktrees (under lock, from #425), and falls back to commit worktrees on ref mismatch. Pre-flight checks were redundant.

**Key evidence**: The `applyAfterFetch` code path already skips pre-flight in apply mode and the full test suite passes. This proves the codebase already trusts apply to self-heal — we're just making that trust explicit for all apply paths.

| Issue | syncMember behavior | Pre-flight needed? |
|---|---|---|
| `missing_bare` | Auto-clones inside `withRepoLock` | No |
| `broken_worktree` | Cleans up + recreates inside `withWorktreeLock` | No |
| `ref_mismatch` | Falls back to commit worktree | No |
| `dirty`/`unpushed` | Skips member unless `--force` | No (warning only) |

## Changes

- `engine.ts`: Pre-flight only runs in lock mode (was: lock + apply with growing exception list)
- `store-hygiene.ts`: `runPreflightChecks` simplified — removes `mode`/`commitMode` params, apply-mode exception lists. All errors block (lock mode is the only caller).
- `pin.ts`: Remove `mode: 'lock'` param (no longer exists)
- Tests: Replace apply-mode-specific tests with `broken_worktree` blocking test

Net: **-55 lines, +44 lines** — simpler code that eliminates the race by construction.

## Known limitation

A deeper clone race remains (#431): `StoreLock`'s `FileSystemBacking` never creates lock files due to missing `FileSystem`/`Path` context propagation. This causes bare repo clone collisions when nested megarepos share secondary members with a clean store. Filed as a separate issue since it's a pre-existing bug in #415, not introduced by this PR.

## Test plan

- [x] All 453 tests pass
- [x] TypeScript type check passes
- [x] Lint/format passes (check:quick)
- [x] Reproduced on dev3 — pre-flight error eliminated, clone race is separate (#431)

### DIFF
--- a/packages/@overeng/megarepo/src/lib/store-lock.ts
+++ b/packages/@overeng/megarepo/src/lib/store-lock.ts
@@ -15,7 +15,8 @@
 
 import { createHash } from 'node:crypto'
 
-import { FileSystem, Path } from '@effect/platform'
+import { NodeFileSystem } from '@effect/platform-node'
+import * as NodePath from '@effect/platform-node/NodePath'
 import { Context, Duration, Effect, Layer, SynchronizedRef } from 'effect'
 
 import type { AbsoluteDirPath } from '@overeng/effect-path'
@@ -46,11 +47,16 @@ export class StoreLock extends Context.Tag('megarepo/StoreLock')<StoreLock, Stor
 type Semaphore = Effect.Effect.Success<ReturnType<typeof DistributedSemaphore.make>>
 
 /**
- * Context needed by both DistributedSemaphore and FileSystemBacking at runtime.
- * FileSystemBacking's operations (tryAcquire, release, etc.) internally yield
- * FileSystem + Path, so these must be available when withPermits runs.
+ * Self-contained layer providing DistributedSemaphoreBacking + FileSystem + Path.
+ * FileSystemBacking operations internally yield FileSystem + Path at runtime,
+ * so both must be available when withPermits runs backing operations.
  */
-type BackingContext = DistributedSemaphoreBacking | FileSystem.FileSystem | Path.Path
+const makeBackingLayer = (lockDir: string) =>
+  Layer.mergeAll(
+    FileSystemBacking.layer({ lockDir }),
+    NodeFileSystem.layer,
+    NodePath.layer,
+  ).pipe(Layer.provide(Layer.merge(NodeFileSystem.layer, NodePath.layer)))
 
 /**
  * Create a keyed lock function backed by distributed semaphores.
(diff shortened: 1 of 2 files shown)


schickling-assistant:
Superseded by #432 which includes the pre-flight skip plus the in-process gate fix for the clone race.

PR #553, opened 2026-04-10
fix: default local pnpm tasks to a workspace pnpm home

schickling:
## Why

Local `devenv` pnpm tasks were still inheriting the global pnpm home when `PNPM_HOME` was unset. That let stale GVS projections from unrelated repos leak into the task state and diverge from the repo-local install model we already use in CI and Nix builds.

## What

- default shared pnpm tasks to `${config.devenv.root}/.pnpm-home` when `PNPM_HOME` is unset
- apply that default consistently across install, update, clean, status, and shell entry
- extend the shell and smoke tests to verify the default and preserve explicit overrides

## How

The shared helper now sets a workspace-local `PNPM_HOME` only when callers have not configured one already. That keeps the projection path stable and local without changing explicit caller configuration.

## Rationale

This keeps the source of truth in the shared pnpm task layer instead of adding downstream workarounds in dotfiles/forge. It reduces cross-repo drift, matches the CI/runtime assumptions we already have, and gives downstream Electron packaging a reliable install boundary.

Acting on behalf of the user.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,7 +70,7 @@ jobs:
           path: |
             ${{ github.workspace }}/.pnpm-home
             ${{ runner.temp }}/pnpm-store/${{ github.job }}
-          key: "pnpm-home-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('**/pnpm-lock.yaml') }}"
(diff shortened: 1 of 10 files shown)


schickling:
Closing this in favor of #559.

The original branch mixed three concerns that no longer belong together:
- CI cache work that has already landed via #556
- unrelated `pty-effect` changes
- the one remaining valid pnpm-home default change

#559 carries only the still-relevant `PNPM_HOME` default behavior on top of current `main`, which is the cleaner branch to review and merge.

_Acting on behalf of the user._

PR #564, opened 2026-04-14
feat(genie): add packages option to netlifyStorybookCommentStep

schickling-assistant:
## Summary

- Add optional `packages` parameter to `netlifyStorybookCommentStep` for repos with non-default storybook layouts
- When provided, generates explicit per-package checks instead of globbing `packages/@overeng/*`
- The `{ path, name }` shape matches the existing Nix `taskModules.netlify` / `taskModules.storybook` config
- Backward compatible — existing callers (effect-utils) pass no opts and use the default glob

## Rationale

The dotfiles repo stores storybooks at `flakes/*/storybook-static` with Netlify aliases like `flakes-oi`, which doesn't match the hardcoded `packages/@overeng/*` glob. Rather than adding fragile glob+strip parameters, accept the same `{ path, name }` package list already used by the Nix task modules.

### DIFF
--- a/genie/ci-workflow.ts
+++ b/genie/ci-workflow.ts
@@ -1056,13 +1056,42 @@ export const netlifyDeployStep = () => ({
 
 /**
  * Combined deploy comment step for Netlify storybook previews.
- * Discovers deployed storybooks by scanning for `storybook-static` build output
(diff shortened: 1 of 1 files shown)


schickling-assistant:
Closing — upstream main already refactored the deploy step to extract package metadata from Netlify deploy output (no glob scanning needed).

PR #578, opened 2026-04-16
fix(pty-effect): eliminate followEvents race condition (#577)

### LINKED ISSUE #577: pty-effect: client test 'exposes get, tag mutation, gc, and live event following' times out on macOS CI
## Problem

The pty-effect client test `exposes get, tag mutation, gc, and live event following` consistently times out (30s) on the macOS ARM CI runner after the pty 0.8 client API adoption (#572).

## CI evidence

From run [24497545177](https://github.com/overengineeringstudio/effect-utils/actions/runs/24497545177) (`test (namespace-profile-macos-arm64)` job):

```
× PtyClient > exposes get, tag mutation, gc, and live event following  30044ms
  → Test timed out in 30000ms.
```

All other pty-effect tests pass (7/8 in `client.test.ts`, all 16 in `PtySession.test.ts`, all 3 in `client.unit.test.ts`).

## Context

- Appeared after #572 (adopt pty 0.8 client API)
- Related to but distinct from #557 (Linux reconnect flake in PtySession)
- Test location: `packages/@overeng/pty-effect/src/client.test.ts:200`

---
*Filed by Claude on behalf of @schickling*

schickling-assistant:
## Summary

- Fixes the flaky 30s timeout in the *exposes get, tag mutation, gc, and live event following* test on macOS ARM CI (issue #577)
- Replaces `Stream.asyncScoped` with `Stream.asyncPush` in `followEvents`: `asyncPush`'s `emit.single` is synchronous (returns `boolean`), which is correct for `fs.watch` callbacks — `asyncScoped`'s `emit.single` returns a `Promise` that was being silently discarded
- Simplifies the return type from `Stream.Stream<PtyEvent, PtyError, Scope.Scope>` to `Stream.Stream<PtyEvent, PtyError>` (scope consumed internally by `asyncPush`)
- Updates the test to watch `session_exit` instead of `session_start`: `EventFollower.watchFile` sets its read offset to the current end-of-file when it discovers a new session via `dirWatcher`, so `session_start` (already written at creation) is permanently skipped. `session_exit` fires ~500ms later and is reliably captured. `session_start` tag assertions move to the existing `readRecentEvents` call.

## Rationale

Two distinct races were present:

1. **Wrapper race** — `asyncScoped` defers its setup Effect until the stream is first consumed. `Effect.forkScoped` schedules the consumer fiber but the parent continues; `spawnDaemon` could fire `session_start` before the fiber ever ran. `asyncPush` has the same laziness, but the 500ms window before `session_exit` gives the fiber time to start during `spawnDaemon`'s async gap.

2. **`EventFollower` race** — `scanAndWatchAll` uses `fs.watch` on the session directory; when a new `.events.jsonl` file is detected the callback calls `watchFile`, which opens the file and sets its read offset to the current file size. By then `session_start` is already written and is skipped. This is a limitation of the upstream `@myobie/pty` `EventFollower` — not something to fix here. Watching `session_exit` sidesteps it entirely.

Fixes #577


schickling:
@codex

PR #620, opened 2026-04-23
fix(devenv): preserve symlinks in source CLI wrappers

schickling-assistant:
## Summary
- pass `--preserve-symlinks` and `--preserve-symlinks-main` through `mkSourceCli` Bun wrappers
- keeps source CLIs aligned with symlinked megarepo workspace layouts, where linked packages should resolve through the workspace graph instead of realpaths
- **fix pnpm 11 GVS staleness**: fold `pnpm-lock.yaml` content into `_gvs_hash` so any lockfile change invalidates GVS `links/`, not just `packageExtensions` / `allowBuilds` tweaks

## Verification
- built a `mkSourceCli` wrapper from latest `origin/main` with `nix build --impure --no-link --print-out-paths --expr ...`
- inspected the generated `bin/test-source-cli` wrapper and confirmed both Bun symlink flags are present

## Rationale for the pnpm fix
The first CI on this PR failed `lint` / `typecheck` with `Cannot find package 'fast-check'` inside `effect/.../FastCheck.js`. A rerun failed again with a *different* missing module (`@effect/platform-node-shared`), on a different Namespace runner — same code, same lockfile. Root cause: after #616 bumped the Effect catalog, pnpm 11's GVS `links/` was reused from pre-bump state on Namespace runners whose 50 GB bundled cache persists workspace state across runs. The existing GVS hash only tripped on `packageExtensions` / `allowBuilds` changes, not on lockfile content, so stale symlinks survived the bump. Including the lockfile content in the hash forces a links/ rebuild on any resolution change.

## Downstream
- dotfiles can consume this by bumping `effect-utils`, then remove any local fallback/import workaround for `mkSourceCli`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
+- **devenv/pnpm-install**: Invalidate pnpm 11 GVS links on any `pnpm-lock.yaml` content change, not just `packageExtensions:` / `allowBuilds:` tweaks. Previously, stale GVS symlinks survived lockfile bumps on runners with persistent volumes (e.g. Namespace bundled cache), leading to non-deterministic "Cannot find package/module" errors for transitive deps after catalog updates
 - **@overeng/pty-effect/client**: Fix flaky timeout in `followEvents` (#577) — `asyncScoped`'s setup ran lazily inside the forked consumer fiber, missing events fired before the fiber started. Replaced with `Stream.asyncPush` (setup still lazy, but `emit.single` is now correctly synchronous for `fs.watch` callbacks). Test updated to watch `session_exit` instead of `session_start`, since `EventFollower.watchFile` starts reading at the current end-of-file when a new session is discovered, making `session_start` unreachable via live following.
 
 ### Changed
--- a/nix/devenv-modules/lib/mk-source-cli.nix
+++ b/nix/devenv-modules/lib/mk-source-cli.nix
@@ -23,6 +23,8 @@
 { name, entry }:
 pkgs.writeShellScriptBin name ''
   root="''${WORKSPACE_ROOT:-$PWD}"
+  # Source CLIs in megarepo workspaces depend on linked packages resolving
+  # through the workspace symlink graph, not through realpath locations.
   exec ${pkgs.bun}/bin/bun \
     --preserve-symlinks \
     --preserve-symlinks-main \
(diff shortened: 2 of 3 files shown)


schickling-assistant:
Closing — the approach in this PR has an unresolved conflict with pnpm's isolated `node_modules` layout that can't be fixed without a larger, genie-driven change.

## Diagnosis

Adding `--preserve-symlinks` / `--preserve-symlinks-main` to Bun in `mk-source-cli.nix` broke source-CLI runs non-deterministically (`lint` / `typecheck` jobs here).

- With `--preserve-symlinks`, Bun does not realpath the importing module, so `import X` is resolved by walking up the **logical** source path.
- pnpm's isolated layout stashes transitive deps under `node_modules/.pnpm/<pkg>@<ver>_<peers>/node_modules/`, reachable only *through* the realpath of the parent symlink.
- Walk-up from e.g. `packages/@overeng/genie/node_modules/effect/dist/esm/FastCheck.js` never reaches `.pnpm/effect@3.21.2_xxx/node_modules/fast-check`.
- Non-determinism across reruns (`fast-check` missing one run, `@effect/platform-node-shared/NodeTerminal` the next, `@effect/platform-node-shared/NodeFileSystem` the next) is pnpm's opportunistic root-hoisting: different transitives land at root `node_modules/` depending on install ordering, so walk-up resolution succeeds or fails per runner.
- `test` passed because vitest runs without the flags; `lint` / `typecheck` invoke `genie:run`, which goes through the source-CLI wrapper.

## Design tension

- **Megarepo composition** requires `--preserve-symlinks` so composed cross-repo workspaces are visible at their composed paths rather than realpath'd to their underlying worktrees.
- **pnpm isolation** requires *not* `--preserve-symlinks` so `.pnpm/<pkg>@<ver>/node_modules/` transitives are reachable.

These are genuinely incompatible unless the workspace layout is adapted (via `publicHoistPattern` or elevated direct deps) to satisfy walk-up resolution.

## Path forward (deferred to a follow-up)

A principled fix needs two pieces:

1. **Opt-in flag** in `mk-source-cli.nix`: `preserveSymlinks ? false`. Default off (preserves current behavior). Callers invoked via megarepo opt in.
2. **Genie-enforced contract**: for source CLIs with `preserveSymlinks: true`, genie computes the entry's runtime dep closure and emits either `publicHoistPattern` entries into `pnpm-workspace.yaml` or elevated direct deps on the CLI package. `dt genie:check` fails on drift.

That turns "it works" into a typed contract: preserve-symlinks consumer and pnpm layout stay consistent by construction, no manual hoisting to rot across effect version bumps.

Follow-up issue to be filed.

## Stranded commit

`a61b0729a` (`fix(devenv/pnpm): invalidate GVS links on lockfile content change`) was pushed based on an earlier, wrong hypothesis about GVS staleness. It's harmless but not the fix. Closing this PR will leave the branch behind; no action needed unless we want a clean branch deletion.

PR #773, opened 2026-06-09
feat(devenv/pnpm): per-host shared store + hardlink import (ext4/Linux)

schickling-assistant:
## Problem

The devenv pnpm store is per-worktree (`storeDir: .devenv/pnpm-store-pure-v1`) with `packageImportMethod: clone-or-copy`. On Linux ext4 (no reflink) `clone-or-copy` degrades to a **full copy**, so every worktree/checkout of a repo duplicates its entire dependency set on disk — the dominant disk consumer on Linux dev/build hosts. On macOS APFS the same setting is a cheap native reflink, so the problem is Linux/ext4-specific.

## Change

Make the pnpm store location + import method a **per-host trait** (gated on `pkgs.stdenv.hostPlatform.isLinux`), in the devenv pnpm module (`nix/devenv-modules/tasks/shared/pnpm.nix`):

- **Linux**: one shared content-addressed store at `${PNPM_SHARED_STORE_DIR:-$HOME/.local/share/pnpm/store-shared-v1}` + `--config.package-import-method=hardlink`, so `node_modules` across all worktrees/sub-workspaces hardlink into a single store (marginal cost per worktree ≈ 0 for shared versions). This generalizes the existing `files/`-only shared-files symlink into a fully shared store (the now-redundant files-symlink is skipped on Linux).
- **macOS**: unchanged — keeps `clone-or-copy` + the per-worktree store + the shared-files symlink (native reflink is already cheap; no regression).
- **CI stays job-local** (the existing `[ -z "$CI" ]` gating is preserved — this only affects local dev/build hosts).

The committed `storeDir:`/`packageImportMethod:` policy keys are **not** changed; the store path + import method are driven via env/CLI config so the per-host behavior is purely additive.

## Validation (representative isolated workspaces incl. a native-build dep, pnpm 11.3.0, ext4)

- `node_modules` are hardlinks into the store (link-count 2, shared blob inode).
- A second fresh worktree adds **~0 KB** store delta (cross-worktree content shares inodes).
- Concurrent installs into the shared store are safe (serialized by the existing store-dir `flock`); store stays intact.
- **Global virtual store single-instance / `TS2742` identity holds** — store location is orthogonal to the resolution model (`node-linker=hoisted` + global virtual store); two packages depending on the same dep resolve to one physical instance.
- **Lifecycle-script packages (e.g. node-pty `node-gyp rebuild`) are always copy-imported even under `hardlink`** (link-count 1) — an in-place rebuild provably cannot mutate the shared store for other worktrees. Direct mutation test confirmed. No copy-pinning needed.
- macOS trait verified unchanged via Darwin eval.

## Scope / review note

This is the source-of-truth piece in the shared devenv pnpm module, opened **in isolation** for review. Consumers of this module pick up the per-host trait automatically; a consumer that sets `PNPM_STORE_DIR` itself will need a matching adjustment, which will follow separately once this is reviewed. The change is self-contained and safe to evaluate standalone.

Cutover note: `store-shared-v1` is a fresh path, so the first install per host repopulates it and the old `~/.local/share/pnpm/shared-files` blob can be GC'd; existing per-worktree `.devenv/pnpm-store*` can be cleaned to realize the disk reclaim.

### DIFF
--- a/nix/devenv-modules/tasks/shared/pnpm.nix
+++ b/nix/devenv-modules/tasks/shared/pnpm.nix
@@ -50,11 +50,36 @@ let
       "${config.devenv.root}/.devenv/pnpm-home"
     else
       "${config.devenv.root}/.devenv/pnpm-home/${workspaceCacheName}";
-  defaultPnpmStoreDir =
+  # Per-host store-location trait (fleet decision 0002).
+  #
+  # On ext4/Linux hosts (dev3, dev4) `clone-or-copy` degrades to a full copy
+  # because the root NVMe has no reflink, so every worktree duplicates the JS
+  # dependency closure (~750G measured fleet-wide). There we move the store to
+  # one shared host-level location on the same filesystem as the worktrees and
+  # import packages with `hardlink`, collapsing all per-worktree stores into a
+  # single content-addressed store with ~0 marginal cost per worktree. pnpm
+  # keys `projects/` and `links/` by absolute project path, so a store shared
+  # across worktrees stays correct.
+  #
+  # On APFS/macOS (mbp2025, mbp2021) `clone`/reflink is already cheap natively,
+  # so the bloat is a Linux/ext4 problem only; the trait must not regress mac —
+  # it keeps the per-worktree `.devenv` store + committed `clone-or-copy`.
+  #
+  # CI is unaffected: it sets its own `PNPM_STORE_DIR` (job-local) which always
+  # wins over this default, and the hardlink flag is gated on `[ -z "$CI" ]`.
+  useSharedStore = pkgs.stdenv.hostPlatform.isLinux;
+  # Resolved at shell-eval time so `$HOME` expands per host, not at Nix eval.
+  sharedStoreRootExpr = ''"''${PNPM_SHARED_STORE_DIR:-$HOME/.local/share/pnpm/store-shared-v1}"'';
+  perWorktreePnpmStoreDir =
     if workspaceRoot == "." then
       "${config.devenv.root}/.devenv/pnpm-store-pure-v1"
     else
       "${config.devenv.root}/.devenv/pnpm-store-pure-v1/${workspaceCacheName}";
+  # Shell expression yielding the default store dir for this host. On ext4/Linux
+  # all worktrees + sub-workspaces share one content-addressed store root; on
+  # APFS/macOS each worktree (and sub-workspace) keeps its own `.devenv` store.
+  defaultPnpmStoreDirExpr =
+    if useSharedStore then sharedStoreRootExpr else lib.escapeShellArg perWorktreePnpmStoreDir;
   installTaskName =
     if taskSuffix == null then
       "${taskNamePrefix}:install"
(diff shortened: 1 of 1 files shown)


schickling-assistant:
Additional benchmark evidence on the shared-store trait, from a controlled run on self-hosted ext4 and ZFS Linux runners with two representative Effect-heavy workspaces (~955 and ~603 packages, pnpm 11.8.0, global virtual store on). Three findings that extend the validation in this PR:

### 1. Concurrency correctness — proven at scale under the risky hardlink import

This PR serializes concurrent installs via the store-dir `flock` and asserts the store stays intact. I stress-tested that empirically with the worst case: 3–4 jobs racing to write the *same* content-addressed blobs into one **cold** shared store, **hardlink** import (so `files/` and `links/` share inodes), full integrity flags on:

- a 3× identical-workspace race (maximum blob-write contention) + a 4× mixed cross-repo race (two different repos sharing Effect deps) on ext4, plus a 3× mixed race on ZFS;
- every install `rc=0`; **zero** corruption / race signatures; an independent sweep of **260k+ store blobs found torn=0** (every blob's sha512 == its store name); all consumers `--offline` reinstall coherent (0 downloads).

So pnpm's cafs write-lock + atomic-write-then-rename holds under concurrent writers on both ext4 and ZFS — the `flock` claim is confirmed under maximal contention, not just serialized single writers.

### 2. Reflink is ineffective for the pnpm store on ZFS (hardlink is the only byte win there too)

The PR correctly notes ext4 `clone-or-copy` degrades to a full copy and chose hardlink. Worth recording for anyone tempted to reach for ZFS block-cloning to "dissolve" the hardlink-vs-safety tension: it doesn't. On OpenZFS 2.4.2 with `block_cloning` active, pnpm's clone import produced **0% block sharing** vs copy — proven three ways:

1. cold install: clone `zfs used` == copy (identical bytes);
2. warm `--force` rebuild issued **42,758 FICLONE ioctls** yet still copied ~920 MB more than hardlink;
3. a confound-free `cp --reflink=always` of a fully-committed store tree: **0% sharing**.

The store's many small files defeat ZFS block-cloning's sharing path. Net: hardlink is the only byte win on ZFS as well as ext4 (ZFS then adds ~32% via lz4 compression + marginal bit-rot insurance — not reflink).

### 3. Store integrity is a write-protection problem, not a pnpm-verify or ZFS-checksum problem

A shared **persistent** store adds a threat the per-worktree model doesn't have: in-place blob corruption. pnpm does **not** re-hash existing blobs on read — after an in-place overwrite (same name, new content), `pnpm store status` reports "untouched", and a fresh `--frozen-lockfile` install with verify flags on trusts the store: `rc=0`, 0 re-fetch, 0 integrity error. With hardlink import the corrupted `files/` blob shares its inode with the `links/` file every consumer executes, so corruption propagates to running code, undetected. Two consequences worth baking into the design:

- pnpm's own integrity flags verify **on write, not on read** — they are not the backstop for a shared persistent store.
- The demonstrated threat is a *write actor* (a process overwriting a blob). **ZFS checksums / scrub do NOT catch this** — the filesystem computes a valid checksum for the new bytes; ZFS defends only *bit-rot*. So the real, filesystem-independent integrity gate is **write-protection**: a dedicated store owner + sticky-bit so only the store manager writes. This is load-bearing on ext4 and ZFS alike.

Net recommendation from the run: import = hardlink; FS = ext4-or-ZFS (ZFS buys ~32% space + minor bit-rot insurance, non-gating); the one required correctness gate is write-protection of the shared store. Happy to share the full metrics.


<details>

| field | value |
| --- | --- |
| `agent_name` | 🪻 cl1-orchid |
| `agent_session_id` | 3aff42e1-d902-4ed2-8ea4-60e39f0162e2 |
| `agent_tool_version` | 2.1.202 |
| `agent_runtime` | Claude Code 2.1.202 |
| `runtime_profile` | /nix/store/nc1jjzy8zzwg31q8dig26wl72jwwbgdv-coding-agent-runtime-profile/share/coding-agents/profile.json |
| `skills_manifest` | /nix/store/f8j6sj5i9i1afq6lycd05sfvpgn8x9wl-agent-skills-corpus/share/agent-skills/manifest.json |
| `worktree` | eu-work/schickling-assistant/2026-07-19-ci-cache-pnpm-v2 |
| `machine` | dev3 |
</details>

PR #779, opened 2026-06-11
Add schema-first OTEL contracts

schickling-assistant:
**Problem**

Otelite now provides a useful capture harness, but trace instrumentation and tests still need to hand-encode span attributes. That leaves room for drift between runtime span data, test expectations, and the Effect Schema contracts used elsewhere in the codebase.

**Goal**

Add schema-first OTEL contracts that make span attributes and `span.label` conventions explicit, reusable, and type checked, while keeping otelite as the capture and inspection backend for e2e trace assertions.

**Decisions**

- Add schema-first OTEL attribute contracts under `@overeng/utils/node/otel-attrs` so runtime instrumentation can share the same contract shape as tests.
- Compile an attribute field plan from Effect Schema AST once, and reject unsafe automatic encodings instead of guessing.
- Require explicit policy for risky encodings such as arrays, nested structs, redacted values, and JSON/string coercion.
- Add `OtelSpan.define`, `OtelSpan.with`, and `OtelSpan.annotate` helpers with an enforced `span.label` contract.
- Extend otelite trace expectations with `trace.expectSpan({ span, match })` and `trace.expectAttributes({ attributes, match })`, reusing the existing matcher stack.
- Keep the Notion CLI fixed-output refresh as a separate mechanical commit so the full local gate can evaluate and complete.

**Verification**

Passed locally:

```bash
CI=1 pnpm --filter @overeng/utils exec vitest run src/node/otel-attrs.unit.test.ts
CI=1 pnpm --filter @overeng/utils-dev exec vitest run src/otelite/trace-expect.test.ts
pnpm --filter @overeng/utils-dev exec tsc -b --pretty false
pnpm --filter @overeng/utils exec tsc --noEmit --pretty false
pnpm exec oxfmt --check packages/@overeng/utils/src/node/otel-attrs.ts packages/@overeng/utils/src/node/otel-attrs.unit.test.ts packages/@overeng/utils-dev/src/otelite/trace-expect.ts packages/@overeng/utils-dev/src/otelite/trace-expect.test.ts packages/@overeng/utils-dev/src/otelite/mod.ts packages/@overeng/utils/src/node/mod.ts packages/@overeng/utils/src/node/otel.ts packages/@overeng/utils/package.json.genie.ts packages/@overeng/utils/package.json
pnpm --filter @overeng/utils exec oxlint src/node/otel-attrs.ts src/node/otel-attrs.unit.test.ts
pnpm exec bun packages/@overeng/genie/bin/genie.tsx --cwd packages/@overeng/utils --check --output ci-plain
devenv tasks run check:all
```

Full gate result: `devenv tasks run check:all` passed.

**Complexity**

This adds a small test/runtime contract layer on top of Effect Schema and the existing otelite expectation helpers. The extra abstraction is justified because it centralizes OTEL attribute encoding policy and makes tests assert the same contract that instrumentation uses.

**Concerns**

- The schema compiler intentionally rejects ambiguous encodings, so some call sites may need explicit policy annotations instead of automatic conversion.
- This does not try to model every OTLP concept. Raw otelite rows remain available for tests that need lower-level details.
- The Notion CLI fixed-output refresh is unrelated to OTEL behavior, but was required for the full local gate to run.

**Friction & bottlenecks**

The first full gate could not evaluate because the Notion CLI pnpm deps fixed-output derivation hash was stale. That was refreshed mechanically in a separate commit before rerunning the full gate.

**Follow-ups**

- Downstream repos can migrate remaining manual otelite span assertions to the new schema-first helpers.
- If link/parentage assertions become common, add typed helpers for those instead of requiring raw row inspection.

**References**

Refs #777.


schickling:
@codex

PR #820, opened 2026-06-20
feat(megarepo): guard worktree archive/reap against in-use worktrees

schickling-assistant:
## Problem

`mr store gc`'s cold-reclaim path decides a worktree is "cold" purely from
absence in the workspace liveness manifests (`store-liveness.ts` /
`classifyStoreWorktreePolicy`). It never checks whether a **live OS process is
currently working inside the worktree**.

This is not theoretical — it caused a real incident. A background store-reclaim
renamed an actively-used worktree into `.archive/` while a coding-agent session
had its shell cwd inside it. On Linux a directory rename is transparent to a
process already in that directory: its cwd silently followed the inode into
`.archive/`, and the subsequent git bookkeeping (detach HEAD + free branch, then
a re-`apply` checkout) stripped the worktree out from under the live session.
`git` started reporting "not a git repository" mid-task and untracked work was
orphaned in the archive.

Even current `main` (with the lossless + grace cold GC from #780) has this gap:
the under-lock veto re-check only re-runs the manifest liveness test. A
freshly-created or repinned-but-unregistered worktree can be absent from every
manifest while a session is actively in it.

## Goal

Before any destructive archive/reap, additionally refuse if a live process has
the worktree (or a descendant) as its cwd or open file handle. "Cold" must mean
*no process is in it*, not just *absent from a manifest*. Conservative direction
throughout: in doubt → keep.

## Decisions

- **Orthogonal veto, not a replacement.** The new check is ANDed with the
  existing manifest liveness veto (keep if EITHER protects the worktree). The
  two reasons stay distinguishable: manifest protection keeps as before; the
  process check keeps with a new `reason: process-in-use` and a dedicated
  `megarepo/store/gc/inuse-veto` span carrying the holding pid/path (the
  attribution the incident lacked).
- **Pure/IO seam, mirroring `store-gc-observations.ts`.** `classifyInUse`
  (pure, unit-tested) takes the already-read `{ pid, path }` list plus an
  explicit `excludePids` set and returns the first matching holder. The
  effectful `readWorktreeInUse` does the `/proc` scan and computes the exclude
  set. Keeping `excludePids` a parameter (not baked into the reader) is what
  lets the classifier be unit-tested AND lets the integration test observe a
  spawned holder.
- **Self/descendant exclusion via PPid walk.** The gc process (and any git
  child it spawns with a cwd inside the worktree) must never self-veto. From
  each in-worktree match we climb its PPid chain; if we reach the gc pid it is
  excluded. Walking up from the rare matches is cheaper than materializing the
  whole process tree.
- **Path boundary.** `candidate === wt || candidate.startsWith(wt + '/')` with
  both sides slash-normalized, so a sibling like `<wt>.archive-old` never
  matches by raw string prefix.
- **Non-Linux → keep (see Concerns).** No `/proc` ⇒ `unknown` ⇒ the caller
  treats it as in-use and keeps.

## Verification

Scoped to the changed package (avoided a full-repo `check:all` to stay light on
shared build/CI capacity).

- `dt ts:check` (workspace `tsc --build`): green.
- `dt test:megarepo` (full package suite): green (exit 0).
- Targeted re-run of the touched/related suites:
  - `store-inuse.unit.test.ts`: 10 passed
  - `store-inuse.integration.test.ts`: 2 passed (spawns a real holder process
    with cwd inside a temp worktree, asserts in-use; a sibling `.archive-old`
    dir reads free; post-kill reads free; and the gc-self-veto case reads free)
  - `store-archive.integration.test.ts`: 8 passed
  - `store-gc-observations.unit.test.ts`: 12 passed
  - `cli/store.integration.test.ts`: 21 passed
- `oxlint` + `oxfmt --check` on all changed files: clean (0/0).

## Complexity

One new lib module with a clean pure/IO split and three small call-site veto
blocks reusing the existing lock + outcome plumbing. No new dependencies (uses
`@effect/platform` `FileSystem.readLink`/`readDirectory`, already used by
`store-liveness.ts`). The new module is justified: the probe is distinct,
testable logic that does not belong inside the gc orchestration.

## Concerns

- **macOS (no `/proc`) behavior change.** On a host without `/proc` the probe
  returns `unknown`, which the caller treats as in-use, so `mr store gc`
  archives/reaps nothing there. This is a deliberate conservative choice
  (keep > delete), but it is a real behavior change for non-Linux gc, not a
  silent no-op. A best-effort `lsof` fallback could restore macOS gc later if
  wanted (see Follow-ups). The incident class lives on the Linux hosts.
- **`--dry-run` does not reflect the in-use veto.** Dry-run short-circuits the
  decision before the lock/probe, so a preview may show `archived` for a
  worktree a real run would keep. The real run is the safe one; the preview is
  advisory.
- Open-fd scanning is implemented but defaulted off; cwd is the load-bearing
  signal that matches the incident.

## Follow-ups

- Optional `lsof`-based fallback so macOS `mr store gc` keeps functioning
  instead of degrading to keep-everything.
- Broader fleet attribution/audit work (kernel-level command audit) is tracked
  separately and out of scope here.

## References

Fixes the worktree-archival incident (RCA captured locally). Related: #780
(lossless + grace cold GC), which this hardens.

<details>

| field | value |
| --- | --- |
| `agent_name` | 🥉 cl1-bronze |
| `agent_session_id` | ddfcde53-424d-4e20-9b32-3374c0698f0c |
| `agent_tool_version` | 2.1.179 |
| `agent_runtime` | Claude Code 2.1.179 |
| `runtime_profile` | /nix/store/wbb5q5n2gbk751hcyr5ndp0zrmar602x-coding-agent-runtime-profile/share/coding-agents/profile.json |
| `skills_manifest` | /nix/store/3i12shfqx3wqzq2di3jy1m7f8fn4prmm-agent-skills-corpus/share/agent-skills/manifest.json |
| `worktree` | effect-utils/schickling-assistant/2026-06-20-2026-06-20-guard-worktree-inuse |
| `machine` | dev3 |
</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,23 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
+- **@overeng/megarepo**: Guard the store worktree archive/reap path against
+  in-use worktrees. The cold-GC liveness signal only checks whether a worktree
+  is referenced by some workspace manifest; it never checked whether a live OS
+  process is currently working inside it, so a background `mr store gc` could
+  rename an actively-used worktree into `.archive/` (its cwd silently following
+  the inode) and strip it out from under a live coding-agent session. New
+  `src/lib/store-inuse.ts` adds a Linux `/proc`-based probe — a pure
+  `classifyInUse` seam plus an effectful reader that scans each pid's cwd (and
+  optionally open fds), excludes the gc process and its descendants via a PPid
+  walk, and degrades to "unknown ⇒ keep" on hosts without `/proc`. The probe is
+  wired as an orthogonal veto into all three destructive sites in
+  `mr store gc` (`archiveWorktree`, `archiveRefMismatchWorktree`, `reapArchive`),
+  under the existing `withWorktreeLock` re-check: if a live process holds the
+  worktree the result is `kept` with reason `process-in-use` and a
+  `megarepo/store/gc/inuse-veto` span recording the holding pid/path for
+  attribution. Conservative direction throughout (in doubt → keep).
+
 - **@overeng/genie**: Make `findGenieFiles` return stable repo-relative
   `.genie.ts` paths, with generation/check orchestration resolving them at the
   filesystem boundary. Discovery tests now assert through a canonical-relative
--- a/packages/@overeng/megarepo/src/cli/commands/store/mod.ts
+++ b/packages/@overeng/megarepo/src/cli/commands/store/mod.ts
@@ -36,6 +36,7 @@ import {
   recordObservations,
 } from '../../../lib/store-gc-observations.ts'
 import { validateStoreMembers, fixStoreIssues } from '../../../lib/store-hygiene.ts'
+import { readWorktreeInUse, type InUseHolder } from '../../../lib/store-inuse.ts'
 import {
   collectStoreLiveSet,
   isPathProtected,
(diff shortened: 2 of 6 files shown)


schickling:
Note: we should make sure this also works reliably on macOS without any gotchas

PR #898, opened 2026-07-07
fix(notion-datasource-sync): notion db sync/track work under packaged Node

schickling-assistant:
## Problem

`notion db sync`, `notion db track`, and `notion db export` were broken under
the packaged Node runtime — they only worked with a backstage
`NODE_OPTIONS=--import tsx` preload. Two independent bugs:

1. **`.tsx` import defect bypasses the plain-progress fallback.** The CLI
   sync-progress path loads the TUI via
   `Effect.promise(() => import('@overeng/tui-react'))` (a `.tsx` module),
   wrapped in `Effect.either` so that a load failure falls back to
   `runWithPlainSyncProgress`. Under packaged Node the dynamic import **rejects**
   (Node can't strip JSX), and `Effect.promise` surfaces a rejection as an Effect
   **defect** — which `Effect.either` does **not** catch. The defect propagated
   past the fallback and the command exited 1 with zero output.

2. **`notion db track` isn't routed by the flake wrapper.** The `notion` wrapper
   routed `db` verbs `init|pull|push|sync|export|status|conflicts|forget|restore|doctor`
   to the Node-with-`node:sqlite` runtime but omitted `track`, so `notion db track`
   fell through to the Bun path and failed closed.

The shipped Nix smoke test only ran `notion db sync --help`, so neither bug was
ever exercised.

## Goal

`notion db sync/track/export` run to completion under the packaged Node binary
with no `NODE_OPTIONS` preload, degrading to plain textual progress when the
optional TUI can't load.

## Decisions

- **Catch the load defect at the decision step, not by converting the import to
  `Effect.tryPromise`.** Both fix the runtime bug, but the injectable-loader +
  `Effect.catchAllDefect` shape is unit-testable without mocking the module
  (vitest transpiles `.tsx` fine, so the real import can't be made to fail in a
  test). The catch is scoped to the load step only — it never wraps the command
  effect, so genuine sync errors still surface rather than being silently routed
  to plain-progress-success.
- **Smoke test greps a phrase unique to the Node entrypoint help**
  (`Packaged Node-backed entrypoint from Nix/devenv`) rather than a shared
  command description, so it genuinely proves `track` reaches the Node runtime
  (the Bun `@effect/cli` help does not emit that phrase).

## Verification

Built `nix build .#notion-cli` and ran the packaged binary in `devenv` with
`NODE_OPTIONS` unset against a tracked SQLite workspace:

- `notion db export . --output …` → exit 0; stderr shows the plain-progress line
  `notion db export complete 100%` (proves the TUI-load defect was caught and the
  plain fallback ran). Before the fix this exited 1 with no output.
- `notion db sync .` → exit 0; full plain progress
  (`pulling → querying → hydrate-row 1..6 → pushing → planning → executor-step → complete 100%`),
  `ok: true`.
- `notion db track <data-source> . --mode local` → exit 0, `ok: true`, replica
  SQLite created; full plain progress. Proves both the routing fix and the
  fallback under a real establish.
- `notion db track --help` routes to the Node runtime (prints the Node help
  block) — this is the assertion added to the Nix smoke test.

Regression test (`@effect/vitest` conventions, no network):
`src/cli/sync-progress-fallback.unit.test.ts` injects a loader that reproduces
the real failure (a rejected dynamic import surfaced by `Effect.promise` as a
defect) and asserts the command completes via plain progress. Confirmed **RED**
(uncaught `FiberFailure`) with the `catchAllDefect` line removed, **GREEN** with
the fix.

## Complexity

No new abstraction or dependency. The change extracts the existing import chain
to a named const, adds an optional `loadTui` parameter (default = the real
loader) as a test seam, and inserts one `Effect.catchAllDefect` line plus one
smoke-test assertion.

## Concerns

None significant. The `loadTui` parameter is test-only surface with a default;
the catch is deliberately narrow (load step only).

## Friction & bottlenecks

- The prior Nix smoke test only exercised `--help`, which is why two real
  runtime bugs shipped. Strengthened here for `track` routing; broader coverage
  of the live replica path under the sandbox remains network-gated.

## Follow-ups

- The wrapper's routed-verb list is hand-duplicated from the CLI command
  definitions in `build.nix`; the same drift that dropped `track` could recur.
  Worth deriving it from a single source of truth (not in scope here).

## References

Found while building a `notion db` demo. Durable fixes only; demo material is
not part of this PR.

<details>

| field | value |
| --- | --- |
| `agent_name` | 💜 cl1-violet |
| `agent_session_id` | 5ce468e2-a077-44ee-a724-bca1fb992e1c |
| `agent_tool_version` | 2.1.202 |
| `agent_runtime` | Claude Code 2.1.202 |
| `runtime_profile` | /nix/store/34v4ljmkjh4sgzax68wwgggmzf69zmfk-coding-agent-runtime-profile/share/coding-agents/profile.json |
| `skills_manifest` | /nix/store/92pwq9mx1byj8nl9an05dyh4baykvf1d-agent-skills-corpus/share/agent-skills/manifest.json |
| `worktree` | notion-db-packaged-node/schickling-assistant/2026-07-07-notion-db-packaged-node |
| `machine` | mbp2025 |
| `tooling_profile` | dotfiles@unknown |
</details>

### DIFF
--- a/packages/@overeng/notion-cli/nix/build.nix
+++ b/packages/@overeng/notion-cli/nix/build.nix
@@ -80,4 +80,34 @@ pkgs.runCommand "notion-cli"
       echo "notion db track smoke test failed (track not routed to node runtime?)" >&2
       exit 1
     fi
+
+    # Real sync-path regression: a `--help` route check cannot catch the actual
+    # defect it guards against. A progress-bearing verb (track) runs through
+    # `runWithCliSyncProgress`, which dynamically imports the `.tsx`
+    # `@overeng/tui-react` TUI. Under packaged Node that import REJECTS (JSX is
+    # not stripped) and surfaces as an Effect *defect*; the top-level handler is
+    # `Effect.tapError` (failures only), so without the `catchAllDefect` fallback
+    # the command dies BEFORE emitting any structured output. `track` is the
+    # cheapest verb that clears argument+context parsing on an empty workspace
+    # and thus reaches the progress wrapper (`sync` fails earlier as
+    # WorkspaceNotTracked, never reaching it). The build sandbox has no network,
+    # so the establish then fails FAST (auth/connection errors are non-retryable)
+    # with a structured `CliErrorEnvelope` — whose presence proves the TUI import
+    # failed soft and real command dispatch was reached. Deleting the
+    # `catchAllDefect` line in runWithCliSyncProgress makes this RED (raw defect,
+    # no envelope); dropping `track` from the wrapper routing makes it RED via the
+    # Bun-runtime guard check below.
+    track_run_dir="$(mktemp -d)"
+    track_run_output="$(NOTION_API_TOKEN=secret_smoke_invalid_000000000000000000000000 \
+      timeout 90 $out/bin/notion db track 11111111-1111-4111-8111-111111111111 "$track_run_dir" --mode local 2>&1 || true)"
+    if ! printf '%s\n' "$track_run_output" | grep -qE '"_tag": "Cli(Error|Result)Envelope"'; then
+      printf '%s\n' "$track_run_output" >&2
+      echo "notion db track sync-progress smoke failed: no structured envelope — the .tsx TUI import defect was not caught (runWithCliSyncProgress catchAllDefect regressed?)" >&2
+      exit 1
+    fi
+    if printf '%s\n' "$track_run_output" | grep -q 'require the packaged Nix/devenv Node-backed runtime'; then
+      printf '%s\n' "$track_run_output" >&2
+      echo "notion db track fell through to the Bun-runtime guard (not routed to the Node runtime)" >&2
+      exit 1
+    fi
   ''
--- a/packages/@overeng/notion-cli/nix/build.nix
+++ b/packages/@overeng/notion-cli/nix/build.nix
@@ -60,12 +60,24 @@ pkgs.runCommand "notion-cli"
   ''
     mkdir -p $out/bin
     makeWrapper ${unwrapped}/bin/notion $out/bin/notion \
-      --run 'if [ "$#" -gt 1 ] && [ "$1" = db ]; then case "$2" in init|pull|push|sync|export|status|conflicts|forget|restore|doctor) shift; exec ${notionDbRuntime}/bin/notion-db-runtime "$@";; esac; fi'
+      --run 'if [ "$#" -gt 1 ] && [ "$1" = db ]; then case "$2" in init|pull|push|sync|track|export|status|conflicts|forget|restore|doctor) shift; exec ${notionDbRuntime}/bin/notion-db-runtime "$@";; esac; fi'
 
     db_output="$($out/bin/notion db sync --help 2>&1 || true)"
     if ! printf '%s\n' "$db_output" | grep -q 'Reconcile an established workspace'; then
       printf '%s\n' "$db_output" >&2
       echo "notion db sync smoke test failed" >&2
+      exit 1
+    fi
+
+    # `track` must also route to the Node runtime (needs node:sqlite). A missing
+    # route falls through to the Bun-backed wrapper, which fails closed. Grep for
+    # a phrase unique to the Node entrypoint's help block ("Packaged Node-backed
+    # entrypoint from Nix/devenv" — see renderCliHelpText); the Bun @effect/cli
+    # help does not emit it, so this specifically proves `track` is routed.
+    track_output="$($out/bin/notion db track --help 2>&1 || true)"
+    if ! printf '%s\n' "$track_output" | grep -q 'Packaged Node-backed entrypoint from Nix/devenv'; then
+      printf '%s\n' "$track_output" >&2
+      echo "notion db track smoke test failed (track not routed to node runtime?)" >&2
       exit 1
     fi
   ''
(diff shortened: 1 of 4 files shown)


schickling-assistant:
Pushed `06a74dc4`: strengthened the Nix smoke from a `notion db track --help` route check to a **real sync-path run**.

The prior smoke never executed `runWithCliSyncProgress` — the exact path where the `.tsx` `@overeng/tui-react` dynamic import rejects under packaged Node as an uncaught Effect defect. The new smoke runs `notion db track <uuid> <tmpdir> --mode local`, which clears arg+context parsing on an empty workspace (unlike `sync`, which fails earlier as `WorkspaceNotTracked`) and reaches the progress wrapper + the real `.tsx` import; in the network-less sandbox the establish then fails fast (auth/connection errors are non-retryable) with a structured `CliErrorEnvelope`, whose presence proves the import failed soft and dispatch was reached.

Verified RED/GREEN against `nix build .#notion-cli`: deleting the `catchAllDefect` line fails the build with "no structured envelope"; dropping `track` routing trips the Bun-guard check.

### Independent end-to-end re-confirmation

The deployed demo binary was failing because the **devenv profile was stale** (predated this PR's commit) — its wrapper omitted `track` routing and its `main.ts` lacked the `catchAllDefect`, which is exactly why `status`/`doctor` (no progress TUI) worked while `sync`/`track`/`export` failed. After `nix build .#notion-cli` + `devenv shell --refresh-eval-cache`, in `devenv shell` with `NODE_OPTIONS` unset:

- `notion db track <ds> . --mode local` completes (full plain progress: preparing → pulling → querying → hydrate 6/6 → complete).
- `notion db sync . --no-materialize-bodies` exits 0; a `sqlite3 UPDATE … Status='Done'` lands in Notion — verified via the API (`select.name` flips to `Done`).
- The sqlite demo E2E passes **3/3** with a real beat-2 Notion screenshot.


<details>

| field | value |
| --- | --- |
| `agent_name` | 💜 cl1-violet |
| `agent_session_id` | 5ce468e2-a077-44ee-a724-bca1fb992e1c |
| `agent_tool_version` | 2.1.202 |
| `agent_runtime` | Claude Code 2.1.202 |
| `runtime_profile` | /nix/store/34v4ljmkjh4sgzax68wwgggmzf69zmfk-coding-agent-runtime-profile/share/coding-agents/profile.json |
| `skills_manifest` | /nix/store/92pwq9mx1byj8nl9an05dyh4baykvf1d-agent-skills-corpus/share/agent-skills/manifest.json |
| `worktree` | effect-utils/schickling/2026-07-07-notion-demo |
| `machine` | mbp2025 |
| `tooling_profile` | dotfiles@unknown |
</details>

packages/@overeng/notion-cli/nix/build.nix
```diff
@@ -60,12 +60,54 @@ pkgs.runCommand "notion-cli"
...
+      --run 'if [ "$#" -gt 1 ] && [ "$1" = db ]; then case "$2" in init|pull|push|sync|track|export|status|conflicts|forget|restore|doctor) shift; exec ${notionDbRuntime}/bin/notion-db-runtime "$@";; esac; fi'
 
     db_output="$($out/bin/notion db sync --help 2>&1 || true)"
     if ! printf '%s\n' "$db_output" | grep -q 'Reconcile an established workspace'; then
       printf '%s\n' "$db_output" >&2
       echo "notion db sync smoke test failed" >&2
       exit 1
     fi
+
+    # `track` must also route to the Node runtime (needs node:sqlite). A missing
```
schickling, line 72: all this added complexity suggests we probably  want to refactor this in a more principled way to avoid the bun vs node complexity

schickling (changes requested):

packages/@overeng/notion-cli/nix/build.nix
schickling-assistant: Agreed that removing the Bun/Node runtime split would be the more principled architectural direction. This PR is intentionally narrower: it repairs the packaged `notion db track` path by adding the missing route and adds failure-capable coverage for the runtime boundary that currently exists. Eliminating that boundary would expand this focused defect repair into a broader CLI/runtime redesign. The PR body already records the immediate structural risk under **Follow-ups**: the routed-verb list is hand-duplicated from the CLI definitions and can drift again, so it should ultimately come from one source of truth. Do you want that architectural refactor to block this repair, or is it acceptable as follow-up work?

schickling-assistant: Thanks — I agree that eliminating the Bun/Node split is the principled direction. This PR is now stale: its branch has been unchanged since July 7 and predates the Effect v4 migration/current CLI work, so I won’t add speculative runtime changes here. I’m treating the architectural refactor as follow-up work for the current code rather than expanding this superseded repair. The migration is already tracked by issue #925; this thread should be revisited there/on the active branch. Resolving this thread.

schickling-assistant: Thanks — I agree that eliminating the Bun/Node split is the principled direction. This PR is now stale: its branch has been unchanged since July 7 and predates the Effect v4 migration and current CLI work (tracked under #925), so I won't add speculative runtime changes here. Treating the architectural refactor as follow-up work on the active branch rather than expanding this superseded repair. Resolving this thread; the concern should be revisited there.

Issue #917, opened 2026-07-15
CI: cold-install pnpm-state flake on Namespace runners amplified by success()-gated cache-save


schickling-assistant:
Downstream consumers on Namespace cloud runners intermittently fail CI on **lock-changing PRs** with a cold-install `node_modules/.bin` materialization failure (`Executable not found in $PATH: vitest`/`madge`), preceded by a determinate-nix daemon self-test error (`cannot connect to socket at /nix/var/nix/daemon-socket/socket`). Observed on livestorejs/livestore#1397 (effect-utils bump). Same commit reruns rotate which jobs pass/fail → confirmed runner-health flake, not code.

## Why it sticks (amplifier)
The generated per-job `Save pnpm state` step is gated:
```yaml
if: ${{ success() && steps.restore-pnpm-state.outputs.cache-hit != 'true' }}
```
On a lock change every job cold-misses the pnpm-state cache. `success()` = whole-job success, so any job whose install completes but whose later step flakes (or whose daemon self-test fails) never seeds the cache → subsequent reruns keep cold-missing → the flake-prone cold path runs every time until a fully-green run happens by luck.

## Two improvements (durable, careful — not urgent)
1. **Safe cache seeding**: seed the pnpm-state cache when the *install itself* verifiably completed (node_modules/.bin populated, .modules.yaml present, install-state/projection hashes written) even if a *later* job step fails — but NOT when the install is incomplete (avoid caching a broken `links/`). This raises seeding probability → fewer flake-prone cold installs on rerun, without caching corruption. Requires a discrete verify-install step to gate on instead of whole-job `success()`.
2. **Daemon/setup retry robustness**: the existing `run-with-nix-gc-race-retry.sh` wraps task runs, not the Nix-daemon self-test / devenv setup phase. Wrap (or retry) the daemon-setup/install phase so a transient `daemon-socket` failure retries instead of failing the whole job.

Root cause of the flake itself is Namespace/determinate-nix daemon availability (external infra); these two changes reduce its blast radius on downstream CI. Refs: pnpm.nix cold-path ~L455-475 / L575-632; genie CI generator `Save pnpm state` step. Filed from the livestore#1397 investigation; relates to the effect-utils consumption/prevention story (frequent-bump + canary).

PR #919, opened 2026-07-16
fix(pnpm): drop redundant ignoreScripts:true that broke bin-shim linking

schickling-assistant:
## Problem

`commonPnpmPolicySettings.ignoreScripts: true` (genie/external.ts) is a blanket switch that disables every dependency lifecycle script. In pnpm 11 GVS it has an **undocumented side effect**: it also suppresses creation of `node_modules/.bin` shims for packages whose subtree carries a build (e.g. `vitest`, `vite`, `madge`, `astro`, `wrangler`). Downstream consumers that resolve tools via `node_modules/.bin` on `PATH` then fail with `Executable not found in $PATH: "vitest"`.

This was surfaced in livestore CI (livestorejs/livestore#1397, the effect-utils bump), where every red job was a bin-shim miss: test-unit (`vitest`), build-examples/perf (`vite`), lint (`madge`), docs (`astro`).

## Root cause (bisected)

The bump added three `pnpm-workspace.yaml` deltas vs the prior policy: `allowBuilds` native flips, `minimumReleaseAge`, and `ignoreScripts: true`. Local bisection (swap each out, reinstall, check `node_modules/.bin/vitest`):
- remove `ignoreScripts: true` → **bins link** ✅
- remove `minimumReleaseAge` → still broken
- `allowBuilds` all-true → still broken

So `ignoreScripts: true` is the sole cause.

## Why it's redundant (allowBuilds already gates)

`allowBuilds` is pnpm 11's real per-package build gate (stored in `.modules.yaml`; `false` = deny + acknowledge, so nothing builds unless explicitly allowed). Decisive proof: with `ignoreScripts` removed and `allowBuilds` kept, deleting **only** the `esbuild: false` line makes pnpm error `[ERR_PNPM_IGNORED_BUILDS] Ignored build scripts: esbuild@…` (esbuild only) — i.e. `allowBuilds: false` actively gates each native. `ignoreScripts` was doing the same gating with a blunter hammer **and** breaking bin-linking.

## Fix

Drop `ignoreScripts: true` from `commonPnpmPolicySettings`. Keep `allowBuilds` (the real gate).

## Validation (this branch, local)

- effect-utils install: **exit 0, no `ERR_PNPM_IGNORED_BUILDS`** — its tree has no build-script package uncovered by the denylist.
- `vitest` bins now **link** in package `.bin` dirs (previously suppressed).
- Gating preserved: effect-utils' `allowBuilds` is all-`false` (natives from Nix), so it still builds **nothing** — this PR is **behavior-neutral** for effect-utils (only bin-linking changes).

## Posture note (for reviewer)

This makes the `allowBuilds` allowlist **live**: `true` entries would now actually build during install (today `ignoreScripts` blanket-blocks them too, so the allowlist's `true` values are dead). effect-utils has **no** `true` entries, so no build-posture change here. Downstream consumers (e.g. livestore) that declare `true` entries for natives they genuinely need built (wa-sqlite, esbuild for vite) will have those become live — intended.

## Follow-ups

- FOD pnpm-deps hashes for packaged CLIs need refreshing (the install output changed — bins now linked); handled by the evergreen FOD closure.
- Upstream pnpm note: `ignoreScripts` suppressing `.bin` shim creation in GVS is arguably a pnpm 11 bug worth reporting.

Refs overengineeringstudio/effect-utils#917.

### DIFF
--- a/genie/external.ts
+++ b/genie/external.ts
@@ -373,7 +373,19 @@ export const commonPnpmPolicySettings = {
   sideEffectsCache: false as const,
   verifyStoreIntegrity: true as const,
   strictStorePkgContentCheck: true as const,
-  ignoreScripts: true as const,
+  // NOTE: `ignoreScripts: true` was intentionally REMOVED here. It is a blanket
+  // switch that disables every dependency lifecycle script, and in pnpm 11 GVS
+  // it has an undocumented side effect: it also suppresses creation of the
+  // `node_modules/.bin` shims for packages whose subtree carries a build (e.g.
+  // vitest/vite/madge/astro), so downstream consumers that resolve tools via
+  // `node_modules/.bin` on PATH get "Executable not found in $PATH". The build
+  // gating we actually want is already provided by `allowBuilds` below (pnpm's
+  // native per-package build allowlist: `false` denies + acknowledges the
+  // build, so nothing builds unless explicitly allowed), which does NOT touch
+  // bin-shim creation. Keeping `ignoreScripts` was therefore redundant for
+  // gating and actively broke bin linking. See overengineeringstudio/effect-utils#917.
+  // This makes the `allowBuilds` allowlist live: `true` entries now actually
+  // build during install (previously ignoreScripts blanket-blocked them too).
   // This dependency refresh intentionally tracks the newest Effect 3 line and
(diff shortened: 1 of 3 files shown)


schickling:
I don’t like this approach.

PR #961, opened 2026-07-21
Add megarepo member resolver enforcing lock/checkout lockstep

schickling-assistant:
Foundation for fixing livestorejs/livestore#1467 (`devenv shell` failing with `error: attribute 'gh-labels' missing` from a clean checkout). Consumer adoption follows in a downstream PR.

## Problem

A repo whose `devenv.nix` imports Nix helpers from a **materialized megarepo member** — rather than only from the pinned flake input — depends on two things agreeing:

- the member's **Nix outputs** (which build the `genie` CLI), and
- the member's **TypeScript generator sources** (the `.genie.ts` files that CLI evaluates)

Reading `./repos/<member>` directly is what guarantees they're the same revision. But it also feeds an **untracked, unversioned directory into Nix evaluation**. If that checkout drifts from `megarepo.lock` — the lock moved, `mr apply` was never re-run — evaluation dies with an opaque error *before the shell can be entered*. The check that would have explained it (`mr:lock-sync-check`) only runs **inside** that shell, so the diagnostic is unreachable exactly when it's needed.

Note that `repos/` is gitignored, so a `git status`-clean checkout says nothing about whether members match the lock.

## Approach

`lib.megarepoMember.resolve` keeps the lockstep property and makes drift explicit:

| state | result |
|---|---|
| member not materialized | pinned flake input |
| tracking worktree | local checkout (deliberate co-development) |
| commit worktree, matches lock | local checkout (lockstep holds) |
| commit worktree, drifted | throw naming both revisions + the remedy |

Modes are distinguished from the worktree name recorded in the member's `.git` file (`gitdir: …/worktrees/<name>`): a 40-char sha means the member is pinned to a commit and **must** match the lock; a branch name means it's deliberately editable, so its revision is expected to differ and is left alone. That keeps `mr config pin <member> --checkout <ref>` co-development working untouched.

Implementation stays pure — only `pathExists` and `readFile`. `readFileType` refuses to inspect paths reached through the member symlink, and Nix has no `readlink`, so the worktree/clone distinction is made by probing for `.git/HEAD`.

Consumers must load the helper from the **pinned flake input**, never the materialized checkout — otherwise resolving a member would require that member to already be resolved.

## Validation

Evaluated against a real materialized member in all three states:

- **matching** (`b5e983b0` == lock) → resolves to the local flake, `devenvModules` present — no behaviour change
- **drifted** (`b5e983b0` vs lock `f007561eb`, i.e. the reported failure) → throws, naming both revisions and pointing at `mr apply`
- **tracking** (worktree name `main`) → correctly classified as non-commit mode, left alone

`check:quick` passes.

<details>

| field | value |
| --- | --- |
| `agent_name` | 🌄 cl2-vale |
| `agent_session_id` | d1dbf622-1045-4840-aa99-085f5211d5cf |
| `agent_tool_version` | 2.1.215 |
| `agent_runtime` | Claude Code 2.1.215 |
| `runtime_profile` | /nix/store/r0yyx5vd9frm4d8klx3i3c1930d9k2ix-coding-agent-runtime-profile/share/coding-agents/profile.json |
| `skills_manifest` | /nix/store/1rg1cdw91rx4gris2kh0sdwyd4d39nxf-agent-skills-corpus/share/agent-skills/manifest.json |
| `worktree` | effect-utils/schickling-assistant/2026-07-21-megarepo-member-lockstep |
| `machine` | dev3 |
</details>

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -273,6 +273,13 @@
       # CLI guard helpers: .mkCliGuard for single guards, .fromTasks/.stripGuards for task-driven guards
       lib.cliGuard = { pkgs }: import ./nix/devenv-modules/tasks/lib/cli-guard.nix { inherit pkgs; };
 
+      # Resolve a megarepo member for Nix evaluation while enforcing that a
+      # commit-pinned checkout matches megarepo.lock (see nix/megarepo-member.nix).
+      # Consumers MUST load this from the pinned flake input rather than from the
+      # materialized checkout, otherwise resolving the member would require the
+      # member to already be resolved.
+      lib.megarepoMember = { lib }: import ./nix/megarepo-member.nix { inherit lib; };
+
       # Builder function for external repos to create their own Bun CLIs
       lib.mkBunCli = { pkgs }: import ./nix/workspace-tools/lib/mk-bun-cli.nix { inherit pkgs; };
 
--- a/nix/megarepo-member.nix
+++ b/nix/megarepo-member.nix
@@ -1,0 +1,106 @@
+# Resolve a megarepo member for Nix evaluation, guaranteeing that the
+# materialized checkout matches `megarepo.lock`.
+#
+# WHY THIS EXISTS
+#
+# Repos whose `devenv.nix` imports Nix helpers from a megarepo member (rather
+# than only from the pinned flake input) need the member's *Nix* outputs and the
+# member's *TypeScript generator sources* to come from the same revision — the
+# `genie` CLI built by Nix evaluates `.genie.ts` sources read from the same
+# checkout. Reading `./repos/<member>` directly achieves that lockstep, but it
+# feeds an untracked, unversioned directory into Nix evaluation: if the checkout
+# drifts from `megarepo.lock` (e.g. the lock moved but `mr apply` was never
+# re-run), evaluation fails with an opaque error such as
+# `error: attribute 'gh-labels' missing` before the shell can even be entered —
+# and the diagnostic that would explain it (`mr:lock-sync-check`) lives *inside*
+# that shell. See livestorejs/livestore#1467.
+#
+# This helper keeps the lockstep property but makes drift an explicit, actionable
+# error instead of an opaque one.
+#
+# WORKTREE MODES
+#
+# `mr` materializes members in one of two modes, distinguishable from the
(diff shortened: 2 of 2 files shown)


schickling-assistant:
Closing in favour of a simpler fix that removes the problem rather than guarding it.

This PR added an eval-time drift guard so that a repo reading its materialized `repos/effect-utils` in Nix would fail legibly instead of opaquely. It works — verified against all three states (absent / matching / drifted). But further investigation changed the conclusion:

1. **`mr status` already detects this drift**, emitting `syncNeeded` and `syncReasons` ("commit drift: workspace is X but lock records Y"). Nothing consumes it. Shipping this helper would have meant two implementations of one invariant, which can diverge — and Nix eval can't shell out to `mr` without IFD, so the duplication would be permanent.
2. **Only 2 of 8 members read the materialized checkout in Nix at all.** The other six use the pinned input, which is immune to this failure mode.
3. **The local read bought no capability** — the member's flake is the same content as the pinned input. It existed only to force revision agreement, which `mr:lock-sync-check` already guarantees (devenv.lock == megarepo.lock, so the pinned input *is* the locked revision).

So the durable fix is to converge those two repos onto the fleet idiom instead of making a divergent pattern safe: livestorejs/livestore#1470.

The real defect this investigation surfaced is that **`mr apply` can materialize a stale revision despite an unmodified lock** — filed separately. The remaining useful work is to gate the existing `mr status` drift signal in shared checks, which restores a loud failure for the whole fleet without new detection logic.


<details>

| field | value |
| --- | --- |
| `agent_name` | 🌄 cl2-vale |
| `agent_session_id` | d1dbf622-1045-4840-aa99-085f5211d5cf |
| `agent_tool_version` | 2.1.215 |
| `agent_runtime` | Claude Code 2.1.215 |
| `runtime_profile` | /nix/store/r0yyx5vd9frm4d8klx3i3c1930d9k2ix-coding-agent-runtime-profile/share/coding-agents/profile.json |
| `skills_manifest` | /nix/store/1rg1cdw91rx4gris2kh0sdwyd4d39nxf-agent-skills-corpus/share/agent-skills/manifest.json |
| `worktree` | livestore-contrib/schickling/2026-07-19-misc |
| `machine` | dev3 |
</details>

PR #1102, opened 2026-08-23
fix(pnpm): project staged source policy into prepared roots

schickling:
## Problem

Prepared pnpm roots retain staged-source and `file:` policy locators, but their dependency source intentionally excludes writable `.devenv` state. Frozen installs therefore need a contained projection onto the declared immutable source closure without changing workspace policy bytes or activating undeclared package source.

## Goal

Make prepared roots faithfully materialize the declared staged-source and policy-only manifest closure while preserving the manifest-alias authority introduced by #1103.

## Decisions

- Keep #1103's staged locator and manifest-alias semantics as the source of truth; the older logical-locator rewrite and `.devenv`-absence contract are retired.
- Stage retained `file:` policy targets as manifest-only inputs after path containment and package-name validation; do not activate their source or dependencies.
- Project staged sources only for the frozen install, fail closed on missing, escaping, or pre-existing targets, and purge the projection before archival.
- Preserve `packageExtensionsChecksum`; retain `pnpmfileChecksum` only when the referenced pnpmfile is present.
- Package the staging entrypoint and its relative filesystem helper in one Nix store output so generated wrappers retain a complete ESM closure.

## Verification

- PASS: focused seven-check projection/checksum suite.
- PASS: `CI=1 devenv tasks run check:quick` on the final tree.
- PASS: real Nix-built `pnpm:install` wrapper regression plus all 36 pnpm smoke cases.
- PASS: refreshed `CI=1 devenv tasks run check:all` on exact commit `b8e5ca0b623591313a9e70bce7a7164db5dfeed8`; terminal exit 0 and `all checks passed!`.
- PASS: independent semantic/security review and `git diff --check`.

## Complexity

The extra logic remains inside the shared prepared-root boundary. Consumers keep one workspace policy and do not gain project-specific locator rewriting.

## Concerns

- Policy-only targets intentionally stage `package.json` only; future pnpm policy fields that inspect additional files must extend this explicit contract.
- Install-time projections are temporary and must remain coupled to the prepared-state cleanup gate.

## Friction & bottlenecks

- Rebase exposed a real authority conflict between the old locator rewrite and #1103's manifest aliases; the stack now uses one authority instead of attempting to preserve both.
- The first refreshed full gate was stopped below the 50 GiB floor; the serialized retry passed after managed Nix GC restored capacity.

## Follow-ups

- Repin schickling/dotfiles PR #1964 to this stack after merge and refresh its measured downstream fixed-output hashes with Evergreen.

## References

- Depends on #1103.
- Related: schickling/dotfiles#1964.
- Related incident: schickling/dotfiles#1880.

### DIFF
--- a/nix/devenv-modules/tasks/shared/pnpm.nix
+++ b/nix/devenv-modules/tasks/shared/pnpm.nix
@@ -104,9 +104,12 @@ let
   nodeModulesProjectionScript = pkgs.writeText "check-node-modules-projection-health.cjs" (
     builtins.readFile ./check-node-modules-projection-health.cjs
   );
-  stageSourceInputsScript = pkgs.writeText "stage-pnpm-source-inputs.mjs" (
-    builtins.readFile ./stage-pnpm-source-inputs.mjs
-  );
+  sourceInputStagingScripts = pkgs.runCommand "pnpm-source-input-staging" { } ''
+    mkdir -p "$out"
+    cp ${./stage-pnpm-source-inputs.mjs} "$out/stage-pnpm-source-inputs.mjs"
+    cp ${./source-input-staging-fs.mjs} "$out/source-input-staging-fs.mjs"
+  '';
+  stageSourceInputsScript = "${sourceInputStagingScripts}/stage-pnpm-source-inputs.mjs";
   sourceStagePath = ".devenv/pnpm-source-inputs";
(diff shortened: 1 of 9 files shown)


schickling-assistant:
Closing as superseded by merged manifest-boundary handling and the Buck2 build-authority cutover. Current dotfiles resolves merged effect-utils main rather than this branch. The branch history remains available if a narrowly scoped fix is later proven necessary.

PR #1105, opened 2026-08-24
fix(ci): anchor helper sources to workspace

schickling:
## Summary

- resolve checked-in Genie CI helpers from `${{ github.workspace }}`
- keep the prepared runtime rooted in the same explicit workspace
- assert generated consumers preserve the absolute source path and setup ordering

## Why

A dotfiles workflow-dispatch proof at exact head `fb13b0bcc039d292c7e94028ccd7d9a8f46da2b6` showed the checked-in `genie/ci-scripts` directory could not be found on custom runners even though it was present in the checkout. The runner shell's implicit working directory is not a stable ownership boundary; `github.workspace` is.

## Verification

- `git diff --check`
- focused Vitest: pending hosted CI (local pnpm environment is unavailable and the host is below its Nix capacity floor)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+
+- **Genie CI workflows**: resolve checked-in helper scripts from the explicit
+  GitHub workspace root so custom runner working directories cannot hide them.
+
 ### Added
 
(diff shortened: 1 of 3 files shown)


schickling:
Closing after deeper failure-log analysis: helper staging and resolve-devenv.sh both succeeded. The missing artifact is prepare-job-local-rust-state.sh, which the downstream consumer never generated. Fixing that source-completeness gap downstream instead of retaining this unrelated path change.

PR #1170, opened 2026-09-01
refactor(stylex): browser-pure token package, virtual-module CSS injection, story gate

schickling-assistant:
## What

PR 1 of the StyleX migration of `@overeng/geist-design-system` — the design system behind `vista/{blocks,service,cli}`, `factory`, `service-hub` and `op-proxy`.

Three commits, deliberately separated so the token move can be shown to change nothing:

1. `d17edc7` — pin the shared layer, adapt to its rename, stand up the story gate. No token or component change.
2. `a5ee269` — fix five references to tokens that were never defined. Changes rendering, deliberately.
3. `0411dd6` — move all 102 `--ds-*` tokens to StyleX variables and the theme switch to a class. **Zero component files changed.**

## Pin

`effect-utils` pinned to `schickling-assistant/2026-09-01-stylex-shared-layer` @ **`6f4cf0518c2f9dc42746f8610124b73802db818e`** in `megarepo.lock` **and** `devenv.yaml`/`devenv.lock` — the repo's `mr:lock-sync-check` requires both to agree. That branch is a clean 14-commit fast-forward from this repo's previous pin (`main@37360aa`), verified by merge-base.

The StyleX lint rules (effect-utils#1172) are **not** in this pinned ancestry, so the no-raw-colour and outline-only-for-focus-visible rules were **applied by review, not enforced by lint**, in this PR.

## Proof — all 694 utility references still resolve, in both schemes

Measured, not reasoned. Both refs were built through the **real pipeline** (Vite + `@tailwindcss/vite` + `createStylexVitePlugins`, LightningCSS-minified) and then loaded into headless Chromium, where every value was read back with `getComputedStyle`.

The 694 source-level `var(--ds-*)` occurrences in `src/components` compile to **194 distinct emitted rules** carrying **197 `var(--ds-*)` occurrences** across **208 declarations** (Tailwind dedupes the repeats). Every one of those was measured on both sides.

**Static — the consuming rules are untouched**

| | before `a5ee269` | after `0411dd6` |
| --- | --- | --- |
| rules referencing `--ds-*` | 196 | 196 |
| byte-identical across the two | 194 | 194 |
| the only two that differ | `:root,[data-theme=dark]` and `[data-theme=light]` | `:root,.x10tp16z` and `.x11l64xu.x11l64xu,.x11l64xu.x11l64xu:root` |

Only the declaration sites moved. Post-minification the 102 dark and 26 light declarations are byte-identical on both sides — zero value diffs, zero missing, zero extra.

**Browser-measured — computed values**

| scheme | tokens compared | token diffs | tokens resolving to nothing | declarations compared | declaration diffs |
| --- | --- | --- | --- | --- | --- |
| dark | 102 | **0** | 0 | 208 | **0** |
| light | 102 | **0** | 0 | 208 | **0** |

`color-scheme` measured as `dark` / `light` correctly on both sides; `getTheme()` and `getResolvedTheme()` agree on both sides.

**The `system` bug, reproduced and fixed.** Chromium with `prefers-color-scheme: light`, `setTheme('system')`:

| | `--ds-background-100` | `color-scheme` | `getResolvedTheme()` |
| --- | --- | --- | --- |
| before | `#0a0a0a` (dark) | `dark` | `light` ← disagrees with what is rendered |
| after | `#fff` | `light` | `light` |

And the new live subscription was verified independently: flipping the emulated preference while in `system` mode moved the root class to the dark scheme class and `--ds-background-100` to `#0a0a0a`, alongside a control listener confirming the media-query change event fired.

## Compiler facts established against the pinned 0.19 plugin

- Literal `--`-prefixed keys in `defineVars` emit the property **byte-identically** and evaluate to `var(<name>)` in JS. **This is per KEY, not per file** — a camel-cased key sitting in the same object still hashes. One mistyped key would silently opt a single token out of the bridge while its neighbours kept working.
- `createTheme` with literal keys works; through the unplugin it emits `.<hash>.<hash>, .<hash>.<hash>:root`, comfortably above the variables' `:root, .<group>`. (The bare babel plugin emits the single-class form `.<hash>, .<hash>:root` — the unplugin doubles it.)
- `defineConsts` emits **no CSS at all** — 18 constants produced 18 zero-length rules. That is why the bridge tokens must be variables: nothing outside StyleX can reference a constant.

## Corrections to `.experiments/0003`

| record says | measured at full scale |
| --- | --- |
| 815 classNames | 533 `className=` sites |
| 664 arbitrary-value references | **694** `var(--ds-*)` occurrences in `src/components` (697 before the dangling fixes) |
| 22 files use `data-[focus-visible]`, 42 use `data-[selected]` | those are **occurrence** counts, not file counts — 22 and 44 occurrences across 6 and 13 files |
| three dangling tokens | **five** — `--ds-yellow-700` and `--ds-cyan-700` are also referenced and undefined, in `stories/Gauge.stories.tsx`. The record measured `src/` only, and the gate now covers stories. |
| `colors.ts` has zero importers | it has **one**, `tailwind-preset.ts` — which itself has none. Both deleted. |
| 102 tokens | 102 `--ds-*`, plus **18** more locally-defined custom properties (`--font-*`, `--radius*`, `--space-*`, `--form-height-*`). Those deliberately stay in `styles.css`: three of them collide with Tailwind's own `@theme` defaults and currently win by source order, so re-emitting them from StyleX would move that collision into a differently ordered layer for no gain. |

One thing the record understated: `--ds-focus-ring` in the light block is byte-identical to the dark one, so 26 light overrides are really 25 plus a no-op. It is kept, because its two nested references re-resolve inside the theme scope either way.

## R08 — the gate is wired but **not yet load-bearing**. Not claiming coverage.

The gate runs end to end here: derived worktree, borrowed `node_modules`, compiler and Tailwind registered in its own Vitest config, one project per theme, ref-derived baseline paths. But reading the baseline report rather than the summary line:

```
942 assertions (471 stories x 2 theme projects)
234 passed, 708 failed
668  "Could not capture a stable screenshot within 5000ms"
 40  genuine pre-existing accessibility violations
 38 of 58 story files fail every story in the file
```

`runStoryGate` builds `preExisting` from baseline failures and skips them in the comparison, so with 71% of stories in that set the regression list is empty **by construction**. That is why this PR's acceptance is the computed-style proof above and not a screenshot diff.

Cause is identified and fixed upstream in effect-utils#1170: transitions firing once on mount. I ran the discriminating A/B here — `Button.stories.tsx`, one theme project, 10 stories: accessibility addon **on** → 10 failed / 10 stability timeouts; addon **off** → identical. So the fix must disable **transitions**, not merely pause animations. This PR will be re-pinned and the gate re-run once that lands.

Note for the next person: every gate invocation ends with `close timed out after 10000ms / something prevents the main process from exiting`. That is a known upstream leak in the StyleX bundler plugin (file handles plus an unref'd timer). Not ours; do not investigate it from scratch.

## Integration work the pin forced

`@overeng/stylex-preset` became `@overeng/stylex-tokens` and the Vite integration moved to `@overeng/utils/node/stylex`. Five references in this repo, two of which are not visible from inside effect-utils:

- `geist/.storybook/main.ts` imported `createStylexVitePlugin` from the old package path
- `devenv.nix`'s pnpm `preInstall` named the old package in a `--filter`

Neither `meters` nor `geist` actually imported the token package, so both drop the dependency rather than repointing it, and the `--filter` goes with it.

**Component-library shape, worth recording because it inverts the app-shaped guidance.** The spec says registering the StyleX plugin in Storybook is redundant because the builder merges the app's `vite.config.ts`. This package has no `vite.config.ts` at all — its only build surface *is* the Storybook preview — so the plugins must be registered in `.storybook/main.ts`, and the preview is the entry named in `entries`.

**Operational trap found:** `mr apply` re-materialises the `repos/effect-utils` store worktree and takes its `node_modules` with it, while `pnpm:install`'s status check does not notice. The symptom is `ERR_MODULE_NOT_FOUND: '@storybook/addon-vitest'` resolved from the effect-utils path — which looks like a missing dependency and is actually a wiped tree. Re-running the `preInstall` restores it.

## Not in this PR

Components (694 references still on utility classes), the rename to hashed names with its eight hand-converted alias tokens, and `@overeng/beads`.

One known deviation to settle with the component work: `stories/StyleXFoundations.stories.tsx` still expresses its focus ring as `boxShadow` with `outline: 'none'`, which is the inverse of the property-partitioning rule. It is left alone here because this PR must be visually neutral.

<details>

| field | value |
| --- | --- |
| `session` | dev3.6xjdfab7 |
| `agent_tool` | OMP |
</details>


schickling-assistant:

<details>

| field | value |
| --- | --- |
| `session` | dev3.mhx73ur2 |
| `agent_tool` | OMP |
</details>

Issue #1171, opened 2026-09-01
Migrate effect-schema-form-aria onto the StyleX enforcement rules


schickling-assistant:
`packages/@overeng/effect-schema-form-aria` does not comply with the StyleX enforcement rules. Each of the 17 non-compliant sites carries a per-line `oxlint-disable-next-line` with a reason. This issue tracks removing them.

## Why this is worth reading rather than just doing

This package is the pilot — the reference implementation the whole StyleX pattern was derived from (migrated in #1152). It now fails the enforcement that pattern produced. That is not embarrassing; it is the enforcement working. The rules found real drift in the one place everyone would have assumed was clean, on their first run, without anyone having to look.

It is also the **second** recorded divergence for this package, not the first. `DELTA-001` (open) records that the pilot expresses accessible-component state by lifting it into JavaScript and selecting whole mutually exclusive style objects, where R05/R06 now prefer the styled element's own conditions with precedence expressed by application order.

**Resolve both together.** They need the same prerequisite — a visual gate that can prove the rewrite did not change rendering — and they touch the same style objects, so doing them separately means paying for the gate twice and reviewing the same diffs twice. `DELTA-001`'s own resolution signal is "the next time the package is substantially modified", which this is.

## Measured

33 diagnostics over 17 sites in four components (`LiteralField.tsx` 8 sites, `NumberField.tsx` 6, `TextField.tsx` 2, `BooleanField.tsx` 1):

- **13 sites** use the deprecated top-level pseudo-class syntax — a condition object at the style-map level rather than nested inside a property. Reported by both `@stylexjs/no-legacy-contextual-styles` (13) and `@stylexjs/valid-styles` (13).
- **4 raw colours** — `#ffffff` x3 in `color`, and `rgb(0 0 0 / 0.1)` twice inside one `boxShadow` value. Reported by `overeng/stylex-no-raw-color` (4) and, for the pure-`color` cases, `@stylexjs/valid-styles` propLimits (3).

The other five enabled StyleX rules already pass on this package. `FieldGroup.tsx` is clean: its `color-mix(in oklab, ${tokens.border} 50%, transparent)` values are derivations over tokens, which follow the colour scheme, and the rule is deliberately silent on them.

Note the `boxShadow` case is the empirical confirmation of experiment 0004: a per-property value limit is keyed on one property, so upstream validation reports nothing about a colour embedded inside that shadow value. Only the first-party rule catches it.

## The tempting wrong fix

[...]

PR #1172, opened 2026-09-01
Fix the oxlint plugin seam, then enforce the StyleX token constraint in two layers

schickling-assistant:
Fixes the oxlint wrapper so third-party JS plugins can load at all, then uses that to ship StyleX enforcement as two complementary layers. Second half depends on the first: without the wrapper fix a config naming both `overeng/*` and `@stylexjs/*` rules fails to resolve, so neither could be tested.

Implements the StyleX spec's "Enforcement" section and decision 0005 Amendment 2.

## Base and merge order — read before merging

Rebased onto `schickling-assistant/2026-09-01-stylex-catalog-pins` (PR #1173, tip `6003e7dd5`), which supplies the `@stylexjs/eslint-plugin` catalog pin this needs. Merge that first.

**The `nix/oxc-config-plugin.nix` `pnpmDepsHash` is contested by construction, not by accident.** Its inputs include the root `pnpm-lock.yaml`, so it moves for any branch that touches dependencies — and this PR adds a devDependency to the very package it covers. #1170 moves it too, for its own reasons. The value here is correct for *this* tree; it is not correct for the merged tree, and neither is #1170's.

So when the second of those two merges: **re-measure from your own `nix build` `got:` line. Do not resolve the conflict by keeping a side, and do not carry a hash across from #1170 or #1173.** A carried-forward hash looks plausible, passes review, and fails after merge, which is the worst time to find out. `evergreen fod refresh` cannot address this attr — it is hand-written, not a CLI-shaped one (tooling gap filed separately); just write what nix reports.

The lockfile move also invalidated two CLI-shaped FODs (`ci-tools`, `notion-cli`), refreshed in the last commit via `evergreen fod refresh --name <pkg> --linux-system x86_64-linux`. Note evergreen writes the correct value even though its post-write verification step fails to evaluate these attrs — the warning is not a failure.

`devenv tasks run check:quick` is **green** on this tree.

### Plugin list replaced instead of substituted — this was the blocker

The wrapper wrote `.jsPlugins = [ourPath]`, discarding every other entry, so a config naming a third-party plugin failed with `Plugin 'x' not found`. Two consumers already carry local workarounds with comments describing exactly this. Now only our entry is substituted.

```
before  $ oxlint fixture.ts --config .oxlintrc.json
        × Plugin 'probe' not found          (exit 1)

after   $ oxlint fixture.ts --config .oxlintrc.json
        × probe(always-report): probe plugin fired
        × overeng(no-raw-nondeterminism): ...   (exit 1, both namespaces)
```

Reproduced in situ, with this PR's real generated `.oxlintrc.json` and the pre-fix wrapper — this is what the blocker actually looked like:

```
$ oxlint packages/@overeng/oxc-config/lt-violation.tsx     # pre-fix wrapper
× Plugin '@stylexjs' not found
```

The match is our **entry point** (`oxc-config/src/mod.ts`), not the package directory, because this PR adds a second plugin entry from that same directory — matching the directory would swap a third-party plugin for ours one level down, reintroducing the same silent failure.

### `--config` first exited 1 with no output

The argument-rewrite loop used `((i++))`. Post-increment evaluates to the *old* value, so it returns status 1 when `i` is 0, and `writeShellApplication`'s `set -o errexit` aborted the wrapper silently. In CI that is indistinguishable from a lint failure.

```
before  $ oxlint --config .oxlintrc.json fixture.ts
        (no output)                          exit 1
        $ oxlint fixture.ts --config .oxlintrc.json
        × probe(always-report): ...           exit 1

after   both orders produce identical output and exit code
```

### A newly added rule was unreachable

The injected plugin is a Nix build-time snapshot, so editing `packages/@overeng/oxc-config/src/*.ts` had no effect and adding a rule reported `not found in plugin 'overeng'`. `OVERENG_OXC_CONFIG_PLUGIN` now overrides the path.

```
before  $ oxlint --config .oxlintrc.json f.ts
        × Rule 'stylex-no-raw-color' not found in plugin 'overeng'

after   $ OVERENG_OXC_CONFIG_PLUGIN=./packages/@overeng/oxc-config/src/mod.ts oxlint ...
        × overeng(stylex-no-raw-color): Raw colour `#00000014` in `boxShadow`. ...
```

### Regression test

`nix/devenv-modules/tasks/shared/tests/oxlint-plugin-injection.test.sh`, 13 assertions, picked up automatically by `devenv tasks run devenv-modules:test`. It drives the **real** built wrapper against the **real** oxlint, because all three defects were failures of plugin *resolution* — the one thing a stub cannot model. It prefers the already-realised wrapper on PATH, so inside the devenv shell it costs nothing and only falls back to `nix build` outside it.

Confirmed to fail against the old wrapper (`Plugin 'probe' not found`) and against an intermediate that matched the package directory rather than the entry point (`Plugin 'sibling' not found`) — it is a regression test, not a tautology.

The two consumer workarounds this unblocks (`devenv.nix` in dotfiles, `flakes/vista/nix/r59-oxc-plugin.js`) are outside this repo; reporting them rather than touching them.

### Why both

Upstream per-property value limits are strictly better than anything we would hand-roll for pure colour properties — they distinguish a literal from a token reference. But a limit is keyed on one property and **structurally cannot see a colour embedded in a composite value**. Measured on this repo's own code:

```
× overeng(stylex-no-raw-color): Raw colour `rgb(0 0 0 / 0.1)` in `boxShadow`.
  LiteralField.tsx:134  boxShadow: '0 10px 15px -3px rgb(0 0 0 / 0.1), ...'
                        ^ upstream valid-styles reports nothing here
```

Banning composite properties outright is not viable — they need literal offsets. The overlap on plain colour properties is kept deliberately: an array `limit` drops its custom `reason` text, so only the first-party message names the remedy.

### Upstream loads through a namespace shim, NOT the bare specifier

`packages/@overeng/oxc-config/src/stylex-upstream-plugin.ts` re-exports the upstream rules with `meta: { name: '@stylexjs' }`.

**Please do not "simplify" this to `jsPlugins: ['@stylexjs/eslint-plugin']`.** It cannot work here, and the reason is structural rather than incidental. Upstream ships no `meta.name`, so oxlint derives the namespace from the specifier — which means the specifier must be the bare package name. A bare specifier resolves only from the **root** `node_modules`, and this repo's root `package.json` is a pure genie workspace aggregate (`{name, private, packageManager, workspaces}`) that cannot carry a dependency at all. Earlier prototyping suggested the bare specifier was required because it ran in an isolated temp project where a root `node_modules` happened to exist.

The shim keeps the dependency on the package that owns lint config, references a path inside the workspace exactly like `./mod.ts` does, and still yields rule names that read exactly as upstream documents them.

### Rules enabled

`valid-styles` (with `propLimits`), `valid-shorthands`, `no-unused`, `no-legacy-contextual-styles`, `no-lookahead-selectors`, `enforce-extension`. `sort-keys` is deliberately skipped — it fights the `property-specificity` resolution the design relies on. Policy lives in `stylexOxlintRules` in `genie/oxlint-base.ts`, repo-wide rather than scoped, so a target is gated the moment it starts using StyleX; the rules are inert in files that never call the API.

`no-lookahead-selectors` is in the set because the spec names it; `no-legacy-contextual-styles` because the brief named it. Shipping the union — both are cheap and neither is a "skip" in either document. It is a good thing: `no-legacy-contextual-styles` is what found the 13 pilot sites.

### Colour properties are enumerated, not globbed

Measured: `*olor*` also matches `colorScheme`, `colorAdjust`, `forcedColorAdjust`, `printColorAdjust` and `colorInterpolation`, all keyword-valued, all of which would be wrongly banned. `fill`, `stroke`, `floodColor`, `stopColor` and `scrollbarColor` are also excluded despite taking colours, because each accepts a value an allowlist would reject (`none`, a `url(#fragment)` paint reference, a pair). The first-party rule still covers colour literals in those.

### `overeng/stylex-no-raw-color`

Bans colour literals as values inside `stylex.create`. The pattern is deliberately **not** anchored, which is what reaches inside composites. The token layer is untouched by construction: only `create` is visited, never `defineConsts`/`defineVars`/`createTheme`.

One refinement beyond the brief, found by running against real code. `color-mix()` and relative-colour syntax (`oklch(from <colour> ...)`) are **derivations**, not constructors — a derivation over a token still follows the colour scheme, so flagging it contradicts the rule's own purpose. Treating them as constructors produced 4 false positives on `FieldGroup.tsx`'s `color-mix(in oklab, ${tokens.border} 50%, transparent)`. They are not a loophole: a raw argument inside them is still caught, because the patterns are not anchored.

Silent on: the token layer, non-colour literals, `colorScheme`/`forcedColorAdjust` keywords, SVG fragment refs (`url(#abc)`) that look like short hex, a quoted hash in `content`, hex runs longer than 8, colour-shaped strings outside `create`, `db.create({ color: '#f00' })`, and colours "forged" across an interpolation boundary. Catches hex 3/4/6/8, seven functional notations, pseudo-class and media-query nesting, pseudo-element blocks, gradients, shadow and border shorthands, template literals, both arms of a conditional in a dynamic style, fallback arrays, and `as const` around the style map.

### `overeng/stylex-outline-focus-visible-only`

Reserves `outline`, `outlineOffset` and `outlineColor` for the focus-visible state. This is the lintable half of the focus-ring invariant: StyleX assigns priority per condition **kind**, not by authoring position, so a selected-state shadow beat a focus-visible shadow and an element that was both selected and keyboard-focused silently lost its focus ring. Partitioning the properties makes the collision impossible rather than managed.

Allows the unconditional base value (so `outline: 'none'` can suppress the UA ring), `default`, at-rule conditions (a forced-colours override is legitimate), and pseudo-elements. Any key containing `focus-visible` counts, covering both `:focus-visible` and the accessible-component library's `[data-focus-visible]`. Plain `:focus` does not — that is the mistake it exists to catch. Fires on either nesting order, once per offending pair, and does not treat dynamic styles as an escape hatch.

### `any` vs TSESTree — a deliberate divergence

The sixteen existing rules annotate the untyped plugin API `any` with a NOTE comment. These two use `TSESTree` from `@typescript-eslint/utils` instead. The no-`any` rule is binding and postdates that convention, the dependency is already there, and the type-only import is erased before bundling — so the cost is zero. The existing sixteen are left alone; churning them is not this PR's job.

### Verification

All of the following was re-run **after** the rebase, on the real generated `.oxlintrc.json` and the real built wrapper, not a hand-written probe config.

Violation fixture — one run, both namespaces resolving:

```
$ oxlint packages/@overeng/oxc-config/lt-violation.tsx
× overeng(stylex-no-raw-color):  Raw colour `#ff0000` in `color`. ...
× overeng(stylex-no-raw-color):  Raw colour `#00000014` in `boxShadow`. ...
× overeng(stylex-no-raw-color):  Raw colour `#fff` in `backgroundImage`. ...
× overeng(stylex-outline-focus-visible-only): `outline` is reserved for the
    focus-visible state but is set under `[data-selected]`. ...
× @stylexjs(valid-styles): color value must be one of: ...
Found 0 warnings and 6 errors.
```

Note what upstream did **not** say: nothing about the colour inside `boxShadow`, and nothing about the colour inside the gradient. Only the first-party rule reaches those. That is experiment 0004's central measurement, reproduced here in situ rather than in a scratch project.

Compliant fixture — both StyleX layers silent, including on all the shapes that must not fire (`color: colors.text` from a `*.stylex.ts` module, `backgroundColor: 'transparent'`, a `color-mix()` derivation over a token, `boxShadow` restyled by `[data-selected]`, `outline` under `:focus-visible`, and `fill: 'url(#brandGradient)'`):

```
$ oxlint packages/@overeng/oxc-config/lt-compliant.tsx lt-tokens.stylex.ts
× overeng(jsdoc-require-exports): Missing JSDoc comment for exported 'const colors'.
× overeng(jsdoc-require-exports): Missing JSDoc comment for exported 'const styles'.
Found 0 warnings and 2 errors.
```

Both remaining diagnostics are an unrelated pre-existing house rule reacting to a throwaway fixture with no JSDoc. Zero StyleX diagnostics.

**Regression test is still non-tautological after the rebase** — checked explicitly, because a rebase across a lockfile change is exactly when that can quietly stop being true:

```
$ OXLINT_WRAPPER_BIN=<post-rebase build>  bash …/oxlint-plugin-injection.test.sh
All oxlint plugin injection tests passed

$ OXLINT_WRAPPER_BIN=<pre-fix wrapper>    bash …/oxlint-plugin-injection.test.sh
FAIL: third-party plugin is not clobbered
  expected output NOT to contain: Plugin 'probe' not found
```

Also:

- 56 unit tests across the two rules, `RuleTester` under both the ESLint and TypeScript parsers. The count matches the enumerated cases exactly, so nothing is silently skipped.
- 13 assertions in the wrapper regression test.
- Repo-wide run with the generated config: **0 errors** over 1537 files, 172 rules. The one warning (`import/no-dynamic-require` in `genie/ci-scripts/bundle-smoke.ts`) is pre-existing.
- `tsc --noEmit` clean for `@overeng/oxc-config`, with all six new files confirmed in the program via `--listFiles`.
- `devenv tasks run check:quick` green.

## 3. The pilot package fails the enforcement derived from it

Enabling the rules produced 33 diagnostics over 17 sites, all in `packages/@overeng/effect-schema-form-aria` — the reference implementation this whole pattern was derived from (#1152).

That is not embarrassing; it is the enforcement working. The rules found real drift in the one place everyone would have assumed was clean, on their first run, without anyone having to look. It is also the **second** recorded divergence for that package, not the first — `DELTA-001` is open for its use of lifted React state where own conditions are now preferred. Both need the same visual gate, so #1171 says resolve them together.

- 13 sites use the deprecated top-level pseudo-class syntax.
- 4 raw colours: `#ffffff` x3 in `color`, `rgb(0 0 0 / 0.1)` twice inside one `boxShadow`.

Not fixed here. Nesting a pseudo-class changes which condition wins on that property, and that package has no visual gate — changing it now would be exactly the unverified visual change our requirements forbid. The colours need semantic tokens that do not exist yet, and the tempting fix (`colors.white`, already in the shared scales) would silence the rule while permanently foreclosing scheme variation, because a constant cannot vary by scheme.

**Each of the 17 sites carries a per-line `oxlint-disable-next-line` with a reason, not a file-scoped override.** A file-scoped `off` would disable the rules for *future* code in that package too — bad anywhere, worse in the package people read as the example. Per-line keeps new code gated, makes the debt visible where it is, and makes the cleanup greppable: `grep -rn 'See #1171'`.

Net effect, stated plainly: the machinery is real and green, but in this repo it currently gates only compliant code. That matches the roadmap's own "Enforcement reach" entry, which expects the third target to be the first lint-covered gate.

## Follow-ups filed

- #1171 — migrate the pilot package onto the rules and remove the 17 disables, together with `DELTA-001`.
- The two consumer wrapper workarounds are now removable; both are outside this repo, so reported rather than touched.

<details>

| field | value |
| --- | --- |
| `session` | dev3.6xjdfab7 |
| `agent_tool` | OMP |
</details>

### DIFF
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -1,7 +1,10 @@
 {
   "$schema": "https://raw.githubusercontent.com/oxc-project/oxc/main/npm/oxlint/configuration_schema.json",
   "plugins": ["import", "typescript", "unicorn", "oxc", "react"],
-  "jsPlugins": ["./packages/@overeng/oxc-config/src/mod.ts"],
+  "jsPlugins": [
+    "./packages/@overeng/oxc-config/src/mod.ts",
+    "./packages/@overeng/oxc-config/src/stylex-upstream-plugin.ts"
+  ],
   "ignorePatterns": [
     "**/node_modules/**",
     "**/.pnpm/**",
@@ -83,7 +86,101 @@
     "typescript/no-unnecessary-template-expression": "off",
     "react/react-in-jsx-scope": "off",
     "react/rules-of-hooks": "error",
-    "react/exhaustive-deps": "warn"
+    "react/exhaustive-deps": "warn",
+    "overeng/stylex-no-raw-color": "error",
+    "overeng/stylex-outline-focus-visible-only": "error",
+    "@stylexjs/valid-styles": [
+      "error",
+      {
+        "propLimits": {
+          "accentColor": {
+            "limit": ["transparent", "currentColor", "currentcolor"],
+            "reason": "Read colours from a semantic token exported by a `*.stylex.ts` module."
+          },
+          "backgroundColor": {
+            "limit": ["transparent", "currentColor", "currentcolor"],
+            "reason": "Read colours from a semantic token exported by a `*.stylex.ts` module."
+          },
+          "borderBlockColor": {
+            "limit": ["transparent", "currentColor", "currentcolor"],
+            "reason": "Read colours from a semantic token exported by a `*.stylex.ts` module."
+          },
+          "borderBlockEndColor": {
+            "limit": ["transparent", "currentColor", "currentcolor"],
+            "reason": "Read colours from a semantic token exported by a `*.stylex.ts` module."
+          },
+          "borderBlockStartColor": {
+            "limit": ["transparent", "currentColor", "currentcolor"],
+            "reason": "Read colours from a semantic token exported by a `*.stylex.ts` module."
+          },
+          "borderBottomColor": {
+            "limit": ["transparent", "currentColor", "currentcolor"],
+            "reason": "Read colours from a semantic token exported by a `*.stylex.ts` module."
+          },
+          "borderColor": {
+            "limit": ["transparent", "currentColor", "currentcolor"],
+            "reason": "Read colours from a semantic token exported by a `*.stylex.ts` module."
+          },
+          "borderInlineColor": {
+            "limit": ["transparent", "currentColor", "currentcolor"],
+            "reason": "Read colours from a semantic token exported by a `*.stylex.ts` module."
+          },
+          "borderInlineEndColor": {
+            "limit": ["transparent", "currentColor", "currentcolor"],
+            "reason": "Read colours from a semantic token exported by a `*.stylex.ts` module."
+          },
+          "borderInlineStartColor": {
+            "limit": ["transparent", "currentColor", "currentcolor"],
+            "reason": "Read colours from a semantic token exported by a `*.stylex.ts` module."
+          },
+          "borderLeftColor": {
+            "limit": ["transparent", "currentColor", "currentcolor"],
+            "reason": "Read colours from a semantic token exported by a `*.stylex.ts` module."
+          },
+          "borderRightColor": {
+            "limit": ["transparent", "currentColor", "currentcolor"],
+            "reason": "Read colours from a semantic token exported by a `*.stylex.ts` module."
+          },
+          "borderTopColor": {
+            "limit": ["transparent", "currentColor", "currentcolor"],
+            "reason": "Read colours from a semantic token exported by a `*.stylex.ts` module."
+          },
+          "caretColor": {
+            "limit": ["transparent", "currentColor", "currentcolor"],
+            "reason": "Read colours from a semantic token exported by a `*.stylex.ts` module."
+          },
+          "color": {
+            "limit": ["transparent", "currentColor", "currentcolor"],
+            "reason": "Read colours from a semantic token exported by a `*.stylex.ts` module."
+          },
+          "columnRuleColor": {
+            "limit": ["transparent", "currentColor", "currentcolor"],
+            "reason": "Read colours from a semantic token exported by a `*.stylex.ts` module."
+          },
+          "outlineColor": {
+            "limit": ["transparent", "currentColor", "currentcolor"],
+            "reason": "Read colours from a semantic token exported by a `*.stylex.ts` module."
+          },
+          "textDecorationColor": {
+            "limit": ["transparent", "currentColor", "currentcolor"],
+            "reason": "Read colours from a semantic token exported by a `*.stylex.ts` module."
+          },
+          "textEmphasisColor": {
+            "limit": ["transparent", "currentColor", "currentcolor"],
+            "reason": "Read colours from a semantic token exported by a `*.stylex.ts` module."
+          },
+          "WebkitTapHighlightColor": {
+            "limit": ["transparent", "currentColor", "currentcolor"],
+            "reason": "Read colours from a semantic token exported by a `*.stylex.ts` module."
+          }
+        }
+      }
+    ],
+    "@stylexjs/valid-shorthands": "error",
+    "@stylexjs/no-unused": "error",
+    "@stylexjs/no-legacy-contextual-styles": "error",
+    "@stylexjs/no-lookahead-selectors": "error",
+    "@stylexjs/enforce-extension": "error"
   },
   "overrides": [
     {
--- a/.oxlintrc.json.genie.ts
+++ b/.oxlintrc.json.genie.ts
@@ -5,17 +5,34 @@ import {
   baseOxlintPlugins,
   baseOxlintRules,
   otelOxlintRules,
+  stylexOxlintRules,
 } from './genie/oxlint-base.ts'
 import { oxlintConfig, type OxlintConfigArgs } from './packages/@overeng/genie/src/runtime/mod.ts'
 
 /** Path to custom oxlint rules plugin */
 const OXC_PLUGIN_PATH = './packages/@overeng/oxc-config/src/mod.ts'
 
+/**
+ * Path to the `@stylexjs` namespace shim, which re-exports the upstream
+ * `@stylexjs/eslint-plugin` rules. A second plugin entry rather than the bare
+ * package specifier, because upstream ships no `meta.name` and a bare specifier
+ * only resolves from the root `node_modules`, which this aggregate root cannot
+ * carry a dependency in. See that file's header.
+ */
+const STYLEX_UPSTREAM_PLUGIN_PATH =
+  './packages/@overeng/oxc-config/src/stylex-upstream-plugin.ts'
+
 export default oxlintConfig({
   plugins: baseOxlintPlugins,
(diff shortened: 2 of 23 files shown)


schickling-assistant:

<details>

| field | value |
| --- | --- |
| `session` | dev3.mhx73ur2 |
| `agent_tool` | OMP |
</details>

PR #1173, opened 2026-09-01
Catalog: Storybook 10.5.10 for per-theme test projects, plus the Tailwind->StyleX converter

schickling-assistant:
Two catalog changes in `genie/external.ts`, both prerequisites for the StyleX migration. They are small in source and large in fallout, so the evidence is below rather than in a follow-up.

Two of the brief's premises turned out to be wrong. Both corrections are in here with measurements, because a corrected record is worth more than a green one.

## 1. Storybook 10.4.6 → 10.5.10

The visual gate needs light and dark covered as **separate Vitest projects**. That is `storybookTest({ initialGlobals })`.

Verified by diffing the shipped plugin option types rather than trusting release notes:

| | `UserOptions.initialGlobals` |
|---|---|
| `@storybook/addon-vitest@10.4.6` `dist/vitest-plugin/index.d.ts` | absent |
| `@storybook/addon-vitest@10.5.10` `dist/vitest-plugin/index.d.ts` | present, documented as *"define one Vitest project per theme"* |

Added in 10.5.0 via [#35226](https://github.com/storybookjs/storybook/pull/35226). Without the bump the gate cannot express the requirement at all.

**10.5.10** is current stable for the whole cohort (`next` is `10.6.0-beta.1`; not taken).

### Breaking changes between 10.4 and 10.5

Upstream `MIGRATION.md` has exactly one entry for `From version 10.4.0 to 10.5.0`: `ExternalDocs` and `ExternalDocsContainer` are deprecated. `grep` finds no use of either in this repo. **Nothing to absorb.**

### Why the whole cohort moves at once

`@storybook/addon-vitest@10.5.10` peers `storybook@^10.5.10`. Pinning the addon without the cohort would fail `strictPeerDependencies`, so this is not a matter of taste.

`@storybook/addon-a11y` is on the list because `parameters.a11y.test` is **inert unless the addon is registered**, and it defaults to warn-only. Pinning the vitest addon alone would have produced a gate that silently passes accessibility regressions.

### Build evidence — one DOM package and one TUI package

The two go through different factories (`createDomStorybookConfig` / `createTuiStorybookConfig`), so one build does not cover the other.

| package | factory | `storybook:build` | resolved | stories in `index.json` |
|---|---|---|---|---|
| `@overeng/effect-schema-form-aria` | DOM (`stylex: true`) | exit 0 | 10.5.10 | 39 |
| `@overeng/tui-stories` | TUI | exit 0 | 10.5.10 | 11 |

Zero `10.4.6` references remain in `pnpm-lock.yaml` or any generated file. (An orphaned `storybook@10.4.6` directory lingers in `node_modules/.pnpm`; zero live symlinks resolve into it — pnpm just has not pruned it.)

### Durable constraint introduced by this PR

**`@vitest/browser`, `@vitest/browser-playwright` and `playwright` are now welded to the `vitest` pin.** `@vitest/browser@4.1.9` peers `vitest` at *exactly* `4.1.9` — a literal version, not a range — and `@vitest/browser-playwright@4.1.9` peers `vitest 4.1.9` plus a **non-optional** `playwright`.

A future `vitest` bump is therefore a **four-package move**, not a one-line change. Whoever hits that next: this is the PR that introduced the coupling. It is also recorded in the catalog comment.

The versions matter, not just the names: `@vitest/browser@4.1.9` ships pixelmatch defaults of `threshold: 0.1` and `includeAA: false` (read from `dist/index.js`), which is the silent-pass trap the gate exists to close.

`playwright` does **not** download browsers at install, on three independent grounds:
1. The published `playwright@1.61.0` tarball's `package.json` has `scripts: null` — no hook exists. Same for `playwright-core@1.61.0`.
2. `ignoreScripts: true` is set fleet-wide in the generated `pnpm-workspace.yaml`.
3. It is **already** in `pnpm-lock.yaml` today, pulled transitively by the existing `@playwright/test@1.61.0` pin. Naming it changes no resolution.

Browsers keep coming from the Nix closure. No `nativeDependencyPolicy` entry needed.

### Reported, not fixed: the deprecated `@vitest/browser/context` import

The brief expected this to "emit a deprecation warning but work via fallback". **The second half is right; the first is not**, and the difference matters.

`@storybook/addon-vitest@10.5.10` `dist/vitest-plugin/test-utils.js:54` imports the legacy `@vitest/browser/context` specifier. But `@vitest/browser@4.1.9` `dist/index.js:855` handles it like this:

```js
if (id === DEPRECATED_ID_CONTEXT) {
  if (importer && !importer.includes("/node_modules/")) {
    globalServer.vitest.logger.deprecate(`... Please, use "vitest/browser" instead.`)
  }
  return VIRTUAL_ID_CONTEXT
}
```

The warning is **suppressed for importers inside `node_modules`**, which the addon always is. So we get **no warning at all** — it resolves silently to the same virtual module. Upstream even comments the constant with *"for libraries that use an older import but are not type checked"*.

Practical consequence: this is **not** noise to live with, it is a **silent Vitest 5 hazard**. The message says the import "will stop working in the next major version", and nothing in our output will remind us. Not fixed here — it is upstream's import, in a dependency, and the addon's Vitest-4 setup file already uses the correct `vitest/browser` specifier, so the blast radius is one call site.

## 2. Tailwind → StyleX converter

Pinned as an **opt-in** dev tool: catalog entry only, no package declares it. A target that wants the first pass picks it up deliberately. This is the mechanism that keeps it away from shared component packages.

### Correction: it is published, and a git-rev pin would not have worked

The brief said to pin an unpublished fork at a git rev. Both halves need correcting.

**It is published.** `tw-to-stylex@0.1.0-alpha.1` on npm, published 2026-02-20 by `mellyeliu` — the same Meta StyleX maintainer — with `gitHead: a150f9e514762c9bab92385e9c0370abda5bc9e6`. That is *the exact fork revision the brief asked me to pin*. We get the intended code through an integrity-pinned npm version instead.

**A git-rev pin would have been broken, not merely policy-blocked.** This is the part worth keeping, because it is the reason not to re-attempt the original approach:

- The fork's `.gitignore` ends with `lib` — the build output is gitignored, so it is in no git ref.
- `package.json` has `prebuild`/`build` but **no `prepare`**, which is the only script pnpm runs for a git dependency.

Measured, twice:

```
$ pnpm add "tw-to-stylex@github:mellyeliu/tw-to-stylex#a150f9e…"      # ignoreScripts: true
[WARN] The git-hosted package … has to be built but the build scripts were ignored.
[WARN] Failed to create bin at …/node_modules/.bin/tw-to-stylex. ENOENT: … 'lib/tw-to-stylex.js'
Done in 4s                                                            # exits 0!

$ node -e "import('tw-to-stylex')"
FAIL: ERR_MODULE_NOT_FOUND Cannot find module '…/tw-to-stylex/lib/index.js'
```

Repeated with lifecycle scripts **fully enabled**: identical result, no `lib/`. There is no `prepare` for pnpm to run, so relaxing `ignoreScripts` fixes nothing. **This was never a dependency-policy question.** Install succeeds, and every invocation dies.

So the answer to "is pinning an unpublished fork acceptable under repo policy" is that the question dissolved: the fork is published, and the git-rev route was not viable under *any* policy.

### The undeclared Babel deps are subtler than "two missing peers"

`@babel/plugin-syntax-typescript` *is* declared as a dependency of `tw-to-stylex`. It still fails, because **Babel resolves plugin names relative to the calling package**, not the plugin's own tree. Reproduced from a clean install:

```
Error: Cannot find package '@babel/plugin-syntax-jsx' imported from /tmp/twbare/babel-virtual-resolve-base.js
# after adding it:
Error: Cannot find package '@babel/plugin-syntax-typescript' imported from /tmp/twbare/babel-virtual-resolve-base.js
```

npm's hoisting masks this half the time; pnpm's strict layout will not. All three (`@babel/core`, `@babel/plugin-syntax-jsx`, `@babel/plugin-syntax-typescript`) are pinned at **7.29.7**. Babel 7, not 8: `@babel/plugin-syntax-typescript@7` peers `@babel/core@^7`, and Babel 8 trips `strictPeerDependencies` (`pnpm peers check` confirms).

Clean install under this repo's actual policy (`ignoreScripts: true`, `strictPeerDependencies: true`) — no warnings.

### Converter evidence, with logging on

`logUnsupported` defaults to `false` in every entry point. Same input, both ways:

**Default (off)** — three classes vanish, zero diagnostics:
```
----- OUTPUT -----
const _styles = _stylex.create({ … })
```

**`{ logUnsupported: true }`**:
```
[TW→StyleX] Unsupported Tailwind pattern: group-*
  Class: group-hover:text-blue-600
  Reason: requires parent element state
[TW→StyleX] Unsupported Tailwind pattern: peer-*
  Class: peer-checked:block
[TW→StyleX] Unsupported Tailwind pattern: [&...] arbitrary selectors
  Class: [&>svg]:size-4
```

It does emit the pattern the spec specifies:
```js
color: { default: "#111827", [_stylex.when.ancestor(":hover")]: "#2563eb" }
// …
<div {..._stylex.props(_stylex.defaultMarker(), _styles.$1)}>
```

### Three traps found while wiring it, all recorded in the catalog comment

1. **The CLI can never log.** `tailwindToStylex()` passes the plugin bare, so `pluginOptions` is `undefined` and `logUnsupported` is permanently `false`. The `tw-to-stylex` binary uses that function. Only `tw-to-stylex/sync` (or the default export used as a Babel plugin *with options*) accepts the flag. **Use of the CLI is ruled out in the comment.**

2. **React Aria `data-[*]` state is dropped with logging ON.** This contradicts the brief, which listed `data-[*]` among what the flag surfaces. It does not — `UNSUPPORTED_CLASS_PATTERNS` only matches `group-`, `peer-`, `*:` and `[&`. Measured:
   ```
   input:  bg-white data-[selected]:bg-blue-500 data-[disabled]:opacity-50 data-[focus-visible]:outline-2
   output: { backgroundColor: "#fff" }        # with logUnsupported: true — zero warnings
   ```
   Three React Aria states gone, silently, with the safety flag on. This is exactly why the spec forbids pointing it at a shared component package, and the flag does **not** make that safe.

3. **The warnings over-report.** It warns that `group-hover:` "will be ignored" and then converts it correctly via `when.ancestor`. Treat the log as a review prompt, not a defect list.

Output also carries raw values (`#111827`, `.5rem`) and leaks Tailwind internals (`--tw-font-weight`), so the hand-pass the spec requires is unavoidable.

## Fallout

`packageExtensions` restated the framework version as a literal `10.4.6` in `genie/internal.ts`, which would have left a second `@storybook/react-vite` resolving. It now derives from the catalog so it cannot drift again.

Changing it moved `packageExtensionsChecksum`, which broke the Nix build **before devenv could enter the shell**:
```
[ERR_PNPM_LOCKFILE_CONFIG_MISMATCH] … "packageExtensionsChecksum" … doesn't match the value found in the lockfile
```
Fixed with `pnpm install --no-frozen-lockfile --lockfile-only`. Worth knowing for anyone touching `packageExtensions`: the repo task cannot bootstrap itself out of this.

Regenerating the lockfile then invalidated **eight** pnpm-deps FODs, cascading one per ~20-minute cycle. Batched instead with a single `nix build --keep-going` over all eight `*-pnpm-deps` attrs, then refreshed and re-verified in one pass. All hashes here are values Nix measured.

**For anyone rebasing onto this branch:** your own dependency changes will invalidate these hashes again. Take the `got:` value Nix prints for *your* build. Do not carry a hash forward from this PR — a carried-forward hash is the classic way these go subtly wrong, because it looks plausible and fails somewhere else.

The `unplugin` duplicate exception was re-checked rather than assumed to be retired by the bump: `@storybook/csf-plugin@10.5.10` still declares `unplugin ^2.3.5`, so it is still required. Its reason now describes the *constraint* rather than naming a consumer package, since the consumer is moving in a sibling PR — and the old text's claim of an "exact pin" was wrong (`^2.3.5` merely resolves to `2.3.11`).

## Tooling gap

`evergreen fod refresh --name <pkg>` writes the correct hash for the seven CLI-shaped attrs, but its **post-write verification always fails** here: it evaluates `import <build.nix> { inherit pkgs; }` while this repo's `build.nix` files require a `src` argument, so every refresh warns `could not evaluate … called without required argument 'src'`. The hash is right; the check is inoperable. I re-verified all eight independently with a batch `nix build`.

`nix/oxc-config-plugin.nix` is worse: `evergreen fod reconcile`/`refresh` cannot address it at all (`could not determine --hash-source for oxc-config-plugin`), and `evergreen fod cli-packages-values` reports **0 known attrs** in this repo. That hash had to be written by hand from the measured value — in a tool we are told to reach for before hand-editing.

## Verification

- `devenv tasks run check:quick` — **green** (exit 0)
- `storybook:build` green for one DOM and one TUI package, both on 10.5.10
- Converter runs from the published package under this repo's pnpm policy, with unsupported-class logging visible
- All eight pnpm-deps FODs rebuild clean

Not run: project-wide formatters or the full test suite, per instructions.

<details>

| field | value |
| --- | --- |
| `session` | dev3.6xjdfab7 |
| `agent_tool` | OMP |
</details>


schickling-assistant:

<details>

| field | value |
| --- | --- |
| `session` | dev3.mhx73ur2 |
| `agent_tool` | OMP |
</details>

PR #1175, opened 2026-09-01
Resolve the pilot package drift the gate, the lint rules and decision 0004 each found

### LINKED ISSUE #1171: Migrate effect-schema-form-aria onto the StyleX enforcement rules

schickling-assistant:
Resolves the pilot package's three independent drift findings. One PR rather
than three because all three were deferred on the same prerequisite — a working
visual gate for this package — which #1170 now supplies. Closes #1171.

## What was wrong, and how each part was proved

| class | fix | gate evidence |
| --- | --- | --- |
| 4 raw colour literals | `on-primary` and `shadow-raised` semantic tokens, defaults from named scale steps | 0 pixels moved |
| 13 legacy top-level pseudo-class sites | nested condition objects; the 2 competing ones moved to ordered `stylex.props` arguments | 0 pixels moved |
| lifted / re-derived React state | React Aria render prop for ancestor-owned state; additive override instead of mutually exclusive objects | 0 pixels moved |
| 10 colour-contrast violations | `primary` `blue500` -> `blue600` | 18 stories, recolour only |
| 1 duplicate-banner violation | `Header` -> plain element + `aria-label` on the group | 0 pixels moved |

All 17 `oxlint-disable` suppressions are gone; all eight rules gate the package
again.

## Gate adjudication

Baseline `98037687a` (#1172 tip): **39 compared, 28/39 passed at baseline, 20
changed.** Non-degenerate, so the pre-existing subtraction is not masking
anything.

- **18 changed are intended.** Every differing pixel is `43,127,255` ->
  `21,93,252`, confined to selected segments, checkbox boxes and the accent
  tick. `blue500` measured 3.76:1 against white body text where AA requires
  4.5:1; `blue600` measures 5.25:1. Renaming the literal to `on-primary` alone
  would have preserved the violation behind a better name.
- **2 changed are the gate's own sub-pixel fringe, not this change.** Proved,
  not assumed: recapturing the *unchanged baseline tree* reproduces both diffs
  identically — `URL` n=689, `With Hint` n=693, same bounding boxes, max channel
  delta 2/255 on one border hairline. Comparing that recapture against this
  branch yields exactly 18 and no fringe. Reported to `GateFixes`, including
  that it is a deterministic **first-capture** effect rather than random noise.
- **0 accessibility failures remain.**

The zero-pixel rows are the substantive result. The recorded hazard was that
nesting a pseudo-class changes which condition wins; the gate shows the
refactor was behaviour-preserving.

## Two things deliberately NOT done

- **The focus ring stays `boxShadow` rather than being converted to `outline`.**
  The partitioning invariant already holds — nothing else on these elements sets
  `outline*`, and `stylex-outline-focus-visible-only` confirms it — and no story
  renders a focused element, so repainting the ring would be a visual change the
  gate structurally cannot adjudicate. **Named follow-up:** add a story that
  renders a focused element, then convert with the gate watching. That turns an
  unverifiable change into a verified one and leaves a permanent regression test
  for a hazard that is otherwise invisible to both the type checker and the
  linter.
- **Two changes are outside the gate's reach and are named as such:** the
  `shadow-raised` token (no story opens the dropdown) and the focus-visible move
  (no story focuses anything).

## Verification

- Gate: as above.
- `lint:check:oxlint`: clean.
- Pilot unit snapshots regenerated (`<header>` gone, `aria-label` present).

Note the reproducible `close timed out after 10000ms / something prevents the
main process from exiting` on every gate run: known upstream StyleX bundler
plugin defect (leaked handles), not ours, recorded so nobody investigates it
from scratch.

<details>

| field | value |
| --- | --- |
| `session` | dev3.6xjdfab7 |
| `agent_tool` | OMP |
</details>


schickling-assistant:

<details>

| field | value |
| --- | --- |
| `session` | dev3.mhx73ur2 |
| `agent_tool` | OMP |
</details>

PR #1176, opened 2026-09-01
Turn on cascade layers for the one target that can take them

schickling-assistant:
Turns cascade layers on for the one target that can take them.

## Why this is a per-target option and not a flat flip

Unlayered output was never a preference — it is what lets converted code beat a
utility framework's layered utilities without any ordering work. So
`createStylexVitePlugins` gains `useCSSLayers`, **defaulting OFF**, and only
`effect-schema-form-aria`'s three configs opt in.

This is a correctness constraint, not taste: the factory is shared through the
megarepo pin, so a flat flip inside it would turn layers on for the other consuming repos
which still carry Tailwind and where layered output loses.

## The audit, in full

- **Utility framework:** none. No `package.json` in the repo names one.
- **StyleX consumers:** exactly one — `@overeng/effect-schema-form-aria` — plus
  the token package itself.
- **Global stylesheets:** five CSS files. Three belong to `notion-react`, which
  does not use StyleX and never shares a document with it here. One
  (`effect-schema-form-aria/src/styles.css`) only re-exports the fifth. The
  fifth is `@overeng/stylex-tokens/preflight.css`.

**"Tailwind-free" turned out not to be sufficient, and that is the finding.**
The reset was unlayered and sets `box-sizing`, `margin`, `padding` and `border`
on `*`. Layered CSS loses to *any* unlayered CSS, so flipping layers on alone
hands those four properties to the reset on every component — silently, and in
exactly the direction this migration exists to prevent. The general rule the
flip actually needs is that **every** unlayered global stylesheet in the graph
is layered and ordered; being free of a utility framework says nothing about it.

## Evidence that the layered output still wins where it must

Both runs against baseline `ff33f7b1b`:

- **Negative control** — reset left unlayered, layers on: the gate fails
  **sixteen stories with DIMENSION mismatches**. Components lose their padding
  and borders and collapse. This is the regression the audit caught.
- **Shipped state** — reset declared in `overeng.reset`, named in the compiler's
  `before` list: **39 compared, 39/39 passed at baseline, 0 changed.** The
  layered output renders identically to the unlayered output it replaces.

## Ordering is fixed by declaration, not by luck

Naming the reset in `before` makes the compiler emit
`@layer overeng.reset, priority1, ...;` ahead of its rules, so the reset sits
below every StyleX priority regardless of which stylesheet the browser parses
first. That is the same "do not depend on bundler injection order" rule the
token layer already follows — this is not another injection-order dependency.

<details>

| field | value |
| --- | --- |
| `session` | dev3.6xjdfab7 |
| `agent_tool` | OMP |
</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,31 @@ All notable changes to this project will be documented in this file.
   rather than a file-scoped override, so new code there is still gated
   (issue #1171).
 
+- **@overeng/utils**: `createStylexVitePlugins` takes `useCSSLayers`, and
+  `@overeng/effect-schema-form-aria` turns it on. Unlayered output was never a
+  preference — it is what lets converted code beat a utility framework's
+  layered utilities without any ordering work — so the option defaults OFF and
+  the flip is per target, on the target that has left the framework. Nothing
+  outside this repo changes.
+  The audit that gates the flip, stated in full because the answer was not the
+  expected one. Utility framework: none — no `package.json` in the repo names
+  one. StyleX consumers: exactly one, `@overeng/effect-schema-form-aria`, plus
+  the token package itself. Global stylesheets: five CSS files, of which three
+  (`notion-react`'s katex, vendored-notion and styles) belong to a package that
+  does not use StyleX and never shares a document with one here, one
+  (`effect-schema-form-aria/src/styles.css`) only re-exports the fifth, and the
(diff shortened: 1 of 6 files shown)


schickling-assistant:

<details>

| field | value |
| --- | --- |
| `session` | dev3.mhx73ur2 |
| `agent_tool` | OMP |
</details>

PR #1183, opened 2026-09-02
test(effect-schema-form-aria): add the focused-control stories the ring conversion needs

schickling-assistant:
Adds the focused-control stories the focus-ring conversion needs. **Additive coverage only — no component or style change.**

## Why this is a prerequisite and not a nicety

Before this file, **no story in the package rendered a focused control.** Every other property of these components is gated; the focus ring was the one thing a conversion could silently repaint with a full pass.

That is not hypothetical. The design system these components borrow their conventions from had a real ring defect ship and survive for years: **8 of 11 ring-carrying files never specified an offset colour**, inheriting the framework's white default, so a keyboard-focused control painted a **solid white 2px halo on a near-black backdrop** — dark mode only. Six internal tools shipped it. It was invisible to a 471-story visual gate because **no story rendered a focused control**, and it was found by reading the cascade by hand after six instruments reported clean.

## Three properties that make these captures usable as baselines

1. **The gate's setup freezes transitions and sets `caret-color: transparent`.** A focused text input otherwise has a blinking caret and no stable baseline at all.
2. **Focus is taken with `userEvent.tab()`, not `element.focus()`.** React Aria decides `data-focus-visible` from its own input-modality tracking, which flips to keyboard only on a key event — a bare `.focus()` focuses the element and paints no ring.
3. **Each `play` asserts the focus state actually landed.** A story that captured an unfocused element would still produce a stable baseline and a green gate. The assertion is what makes the capture evidence rather than a picture.

## What is verified

| claim | evidence |
|---|---|
| Typechecks | in-package `tsc --noEmit -p tsconfig.json`, exit 0 |
| Stories execute | browser launched, **all 4** reached `afterEach` — so all 4 `play` functions ran to completion, focus assertions included |
| Captures are real | 4 reference PNGs, **1776–2832 bytes, four distinct hashes** — not one repeated frame |

The run reports 4 failures. They are `toMatchScreenshot` with **no baseline to compare against**, which is the correct and only possible result for brand-new stories: the first run writes references. **The failures are the absence of a baseline, not a rendering problem.**

**Not claimed:** that the rings look right. That is what the baseline gates from here on.

## Note for anyone running the gate

This must run **inside `devenv shell`**. `PLAYWRIGHT_BROWSERS_PATH` is supplied there and points at the Nix store, which carries exactly the `chromium_headless_shell-1228` the runner asks for. Run outside it and the provider looks in `~/.cache/ms-playwright`, finds nothing, and every story reports as an unhandled browser-launch error with **`no tests`** — which reads like a broken gate rather than a missing environment.

## Provenance

Authored by a delegated agent that parked before committing. Recovered, typechecked, and executed here. Every number above was measured on this branch.

<details>

| field | value |
| --- | --- |
| `session` | dev3.mhx73ur2 |
| `agent_tool` | OMP |
</details>

### DIFF
--- a/packages/@overeng/effect-schema-form-aria/src/stories/FocusRing.stories.tsx
+++ b/packages/@overeng/effect-schema-form-aria/src/stories/FocusRing.stories.tsx
@@ -1,0 +1,145 @@
+import type { Meta, StoryObj } from '@storybook/react'
+import { expect, userEvent } from 'storybook/test'
+
+import { LiteralField } from '../components/LiteralField.tsx'
+import { NumberField } from '../components/NumberField.tsx'
+import { TextField } from '../components/TextField.tsx'
+
+/**
+ * Keyboard-focused controls, one story per element that paints a focus ring.
+ *
+ * Additive coverage, and it is a prerequisite rather than a nicety: before this
+ * file no story in the package rendered a focused control, so the focus ring was
+ * outside the visual gate entirely. Every other property of these components is
+ * gated; the ring was the one thing a conversion could silently repaint with a
+ * full pass. That is the same blind spot that let a real ring defect ship
+ * unnoticed in the design system this package borrows its conventions from.
+ *
+ * Three properties make these captures usable as baselines:
+ *
+ * 1. The gate's setup freezes transitions AND sets `caret-color: transparent`,
+ *    so a focused text input has no blinking caret to make the capture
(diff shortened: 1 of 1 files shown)


schickling-assistant:

<details>

| field | value |
| --- | --- |
| `session` | dev3.mhx73ur2 |
| `agent_tool` | OMP |
</details>

PR #1184, opened 2026-09-02
fix(story-gate): thread consumer plugins into every project, not the root

schickling-assistant:
Top of stack #1177, on #1183. Fixes the reason **the ptg story gate has never executed a story on any pin** — a fact carried across eight PRs and a whole migration with no cause.

## The defect

**Vitest projects do not inherit root-level `plugins`.** Each project is its own Vite config.

`createStoryGateConfig` returns **a bare project config for one theme** and **`{ test: { projects } }` for several**. So the documented consumer pattern —

```ts
defineConfig(mergeConfig(createStoryGateConfig({ themes: [...] }), {
  plugins: [createStylexVitePlugins()],
}))
```

— works **by accident** in the single-theme case, where the merge target *is* the project, and **silently stops applying** the moment a second theme is added. The plugins land beside `test` where nothing reads them, and every story fails to load with a runtime error from untransformed source.

The old comment in each consumer said the plugin "has to be registered here rather than inherited". That was correct, and registering it "here" is exactly what does not work once there is more than one theme. **A correct instruction that cannot be followed reads as a followed instruction.**

## Measured

Same host, same `devenv shell`, minutes apart:

| consumer | themes | shape returned | result |
|---|---|---|---|
| `effect-schema-form-aria` | none | bare project config | **4 stories ran**, 4 distinct references captured |
| `apps/ptg` | 2 | `{ test: { projects } }` | dies at import, **`Tests no tests`** |

**Threading the same plugins per-project moved ptg from `Tests no tests` to 10 stories executed.** The metric that moved is *whether anything ran*, which is all the diagnosis needed.

## The change

- A `plugins` option, applied to **every** project, ordered **before** the Storybook plugin because a compiler transform has to see the source first.
- Storybook plugin construction becomes an **injected seam**. `storybookTest` eagerly loads a real Storybook config directory, and the invariant worth guarding is *where plugins land* — independent of what they are.
- The in-repo consumer migrates off `mergeConfig`.

## The test is proven live, not merely passing

**Reverting the placement to the pre-fix form fails 3 of its 4 cases.** It also:

- asserts **per project**, and separately that the **root does not** carry them — so a refactor cannot satisfy it by duplicating plugins everywhere;
- carries a **negative control** (no plugins supplied) so the threading cannot become load-bearing for consumers needing no transform;
- imports `project.ts` **relatively**, so it is immune to workspace link direction.

## What is not verified, and how I know

**The `effect-schema-form-aria` config migration is unverified.** In my worktree that package resolves `@overeng/utils` to a **sibling checkout** rather than to this source, so its clean typecheck says nothing about this change.

I found that with a **negative control**: a deliberately bogus option produced **no** type error either, which means the check could not have caught a real mistake at that call site. The green was uninformative, not reassuring.

It is a four-line mechanical migration and CI installs properly. Flagging it beats presenting a pass that cannot fail.

## Follow-ups, both filed

1. **ptg's consumer still needs migrating** — separate repo, blocked on this landing.
2. **Blocker 2 is independent and still open.** With the transform applied, all ten ptg stories then throw `Cannot read properties of null (reading 'useState')` — a null hook dispatcher from a third-party component resolving through a sibling worktree's store, because ptg's `node_modules` is a symlink. Duplicate React, same mechanism diagnosed earlier in this migration. **R08 needs both fixes.**

<details>

| field | value |
| --- | --- |
| `session` | dev3.mhx73ur2 |
| `agent_tool` | OMP |
</details>

### DIFF
--- a/packages/@overeng/effect-schema-form-aria/vitest.gate.config.ts
+++ b/packages/@overeng/effect-schema-form-aria/vitest.gate.config.ts
@@ -1,4 +1,4 @@
-import { defineConfig, mergeConfig } from 'vitest/config'
+import { defineConfig } from 'vitest/config'
 
 import { createStoryGateConfig } from '@overeng/utils/node/storybook/gate'
 import { createStylexVitePlugins } from '@overeng/utils/node/stylex'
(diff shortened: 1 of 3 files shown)


schickling-assistant:

<details>

| field | value |
| --- | --- |
| `session` | dev3.mhx73ur2 |
| `agent_tool` | OMP |
</details>

PR #1196, opened 2026-09-03
fix(genie): stop CI generators emitting steps only effect-utils can satisfy

schickling-assistant:
## The defect class

**A CI generator emits something only this repo can satisfy.**

Both bugs here have that shape, and it is worth naming because it explains why this repo's CI never caught either one. `effect-utils` authors the shared CI steps *and* consumes them, but it consumes them through a different path than downstream repos do. Anything that happens to be true locally — a mode that resolves, a script that is hand-committed — makes the generated step work here and fail everywhere else. **Green CI in this repo is not evidence that a generated step works.**

Two instances found and fixed. A third would look like: any generated step referencing state that exists in this checkout but is not part of what a consumer generates or receives.

---

## Instance 1 — implicit `--worktree-mode` in workspace-applying steps

`mr apply` refuses the implicit `auto` worktree mode when `CI=true`:

```
packages/@overeng/megarepo/src/cli/commands/engine.ts
const resolvedWorktreeMode = worktreeMode ?? 'auto'
const appliesWorkspace = mode === 'apply' || applyAfterFetch === true

if (appliesWorkspace === true && resolvedWorktreeMode === 'auto' && process.env.CI === 'true') {
  return yield* new InvalidOptionsError({
    message:
      'CI requires an explicit worktree strategy; pass --worktree-mode commit for deterministic detached worktrees or --worktree-mode tracking for branch worktrees',
  })
}
```

Both generators emitted the invocation without a mode:

| generator | emitted command | trips guard via |
| --- | --- | --- |
| `applyMegarepoLockStep` | `nix run …#megarepo -- apply --all` | `mode === 'apply'` |
| `syncMegarepoWorkspaceStep` | `mr fetch --apply` | `applyAfterFetch === true` |

Every consumer that regenerated got a `Sync megarepo dependencies` step that threw during job setup, before any lint/typecheck/test work ran. Measured blast radius across the two consumer repos: **5** and **14** unflagged `apply --all` sites, none carrying a mode, taking 5 of 6 CI jobs red in the first. This repo: **0** sites — it never applies its own lock this way.

**`applyMegarepoLockStep` → `--worktree-mode commit`.** Behaviour-preserving, which is the decisive argument: before the guard landed, `auto` + `CI=true` resolved to exactly `commit`, so pinning it changes nothing but satisfies the guard. It is also the semantically right mode for a step whose whole purpose is holding the checked-in lock shape — `docs/spec.md` calls `commit` `refs/commits/<sha>/`, "deterministic, isolated — exact locked commit guaranteed".

**`syncMegarepoWorkspaceStep` → `--worktree-mode tracking`.** Different intent, different answer: that step deliberately moves members to latest branch heads, so it wants branch worktrees. Matches the existing choice in `genie/ci-scripts/prepare-effect-utils-composition.sh`, which already passes `tracking` because it needs `mr apply` to attach a job-owned branch ref and asserts `member_ref == branch_ref` afterwards. Zero current consumers — but its only purpose is CI, so it was 100% broken dead API waiting to trap the next adopter.

## Instance 2 — `resolve-devenv-ci.sh` had no support-file entry

Unmasked by fixing instance 1: the failure moved exactly one step down.

`validateNixStoreStepFor` (`genie/ci-workflow/setup.ts:1040`) emits a step invoking `resolve-devenv-ci.sh` out of the prepared runtime scripts dir. That dir is populated by `Prepare CI helper scripts` from the **consuming** repo's `genie/ci-scripts`, which is generated from `ciWorkflowSupportFiles` — and there was no entry for that script. Here it is hand-committed with no generator beside it, so this repo resolves it fine.

Observed on a real runner, not hypothesised. Consumer `lint` job step conclusions:

```
Sync megarepo dependencies   success
Prepare CI helper scripts    success
Resolve devenv               failure   ← exit 127
Format + lint                skipped

.../composition-state/ci-runtime/resolve-devenv-ci.sh: No such file or directory
```

Measured: the consumer's `genie/ci-scripts` holds 6 files; this repo's holds 23, including `resolve-devenv-ci.sh`. A consumer cannot generate what has no entry.

Fixed by adding `resolveDevenvCi`, using the same lazy `sharedScriptArtifact` getter as `prSnapshotArtifact` — this module is reachable from the barrel during genie's cold bootstrap, so the read must stay deferred.

### Why the shared source stays hand-written (deliberate, not an oversight)

It would be tempting to also add a `resolve-devenv-ci.sh.genie.ts` here so this repo's copy is generated too, removing a hand-committed file that has no generator. **That would be worse.** The generator would read the very file it writes — a tautology that can never detect drift, only launder it.

The drift risk it would supposedly close does not exist: consumers emit the script through `repoContext.readText(...)`, which reads **this exact file** out of the effect-utils checkout. Verified on a consumer's generated copy — the only difference is genie's own three-line `Generated file - DO NOT EDIT` provenance banner injected after the shebang, exactly as it already treats the sibling `resolve-devenv.sh`; the script body is emitted verbatim, so the two cannot silently disagree. That is precisely the contract already documented on `prSnapshotArtifact` — "Consumers emit it verbatim, so every repo validates candidates with identical code" — and `pr-snapshot-artifact.mjs` likewise ships hand-written with no `.genie.ts` beside it. Instance 2 was a missing *entry*, not a missing generator.

## Tests — the class, not just the instances

Two new tests in `ci-workflow-helpers.unit.test.ts`, both sensitivity-checked against the pre-fix source rather than assumed to work.

**1. Every workspace-applying invocation carries an explicit mode.** Sweeps `-- apply` and the `'--apply'` argv form across the generator sources; matches command shapes only, never prose.

| source | invocation sites | unflagged |
| --- | --- | --- |
| `origin/main` (pre-fix) | 2 | **2** |
| this branch | 2 | **0** |

**2. Every prepared-runtime script has a support-file entry.** Sweeps the scripts the generators invoke out of the prepared dir, follows the `$script_dir` siblings those scripts source, and fails on any without an entry.

| source | required scripts | uncovered |
| --- | --- | --- |
| `origin/main` (pre-fix) | 5 | **1** — `resolve-devenv-ci.sh` |
| this branch | 5 | **0** |

## Audit: is there a third instance?

Asked and answered — **no**, and the negative result is the point.

Full surface of scripts any generated step invokes out of the prepared runtime dir, plus their transitive `$script_dir` dependencies:

| script | reached via | support-file entry |
| --- | --- | --- |
| `prepare-job-local-rust-state.sh` | `shared.ts:371` | `jobLocalRustState` ✅ |
| `run-with-nix-gc-race-retry.sh` | `shared.ts:407` (`withGcRaceRetry`) | `nixGcRaceRetryWrapper` ✅ |
| `resolve-devenv-ci.sh` | `setup.ts:1040` | `resolveDevenvCi` ✅ *(added here)* |
| `resolve-devenv.sh` | sourced by `resolve-devenv-ci.sh` | `resolveDevenv` ✅ |
| `nix-gc-race-retry.sh` | sourced by the retry wrapper | `nixGcRaceRetry` ✅ |

**5 required, 5 covered, 0 gaps.** Enforced by test 2 above, so instance 3 cannot land silently.

## Verification

- `devenv tasks run genie:run` — clean, **no generated-file changes in this repo**, confirming there is no `applyMegarepoLockStep` site here
- `devenv tasks run test:genie` — **420 tests, 0 failed**, including both new tests
- `devenv tasks run check:quick` — exit **0**

### Liveness: a local green does NOT prove instance 1

Stated plainly because the checklist above invites the wrong conclusion. The guard only fires when `CI=true`, so **no local run exercises that change at all** — `genie:run`, `test:genie` and `check:quick` pass identically with and without the flag. They prove the generator emits it and that nothing else regressed; they cannot prove the fix works.

The load-bearing evidence is a green **`Sync megarepo dependencies`** step in a real CI run of a consumer that has repinned and regenerated. **That evidence now exists:** on runner `dev3-8f72183e`, with `CI: true`, the log shows the executed command as

```
nix run "github:overengineeringstudio/effect-utils/$EU_REV#megarepo" -- apply --all --worktree-mode commit
{"_tag":"Success", …"results":[{"name":"effect-utils","status":"cloned","commit":"1412d4a1a…","ref":"main"}]…}
```

and the step's conclusion is `success`. The consumer's `source-policy` job — the only one that was green before, because it does not use this step — still passes, ruling out a regression there.

Instance 2's fix is verified in-file and by test; its own liveness proof comes from the consumer's next run, which needs this branch merged first.

## Consumers need BOTH fixes

Worth stating explicitly, because it is easy to assume otherwise: merging only instance 1 unblocks nobody. A consumer that repins to instance 1 alone gets past `Sync megarepo dependencies` and then dies at `Resolve devenv`. That is exactly what was observed. One consumer's adoption PR is open and waiting on this branch; the other (14 sites) needs the same repin + regenerate and is **not** done here.

<details>

| field | value |
| --- | --- |
| `session` | dev3.mhx73ur2 |
| `agent_tool` | OMP |
</details>

### DIFF
--- a/genie/ci-workflow/megarepo.ts
+++ b/genie/ci-workflow/megarepo.ts
@@ -102,9 +102,16 @@ fi
   shell: 'bash',
 } as const
 
-/** Fetch latest refs and apply megarepo workspace. */
+/**
+ * Fetch latest refs and apply megarepo workspace.
+ *
+ * `mr fetch --apply` applies the workspace, so CI requires an explicit
+ * `--worktree-mode`. `tracking` matches this step's intent: it deliberately
+ * moves members to the latest branch heads, so it wants the branch worktrees
+ * under `refs/heads/<branch>/` rather than detached per-commit ones.
+ */
 export const syncMegarepoWorkspaceStep = (opts?: { skip?: string[] }) => {
-  const args = ['mr', 'fetch', '--apply']
+  const args = ['mr', 'fetch', '--apply', '--worktree-mode', 'tracking']
   const skipCsv = opts?.skip?.join(',')
   if (skipCsv !== undefined && skipCsv !== '') args.push('--skip', shellSingleQuote(skipCsv))
   return {
@@ -126,6 +133,12 @@ ${args.join(' ')}`,
  * shape instead of silently drifting to newer branch heads during job setup.
  * Resolves the CLI inline with `nix run` to avoid `nix profile install`
  * conflicts on self-hosted runners.
+ *
+ * `--worktree-mode commit` is mandatory: `mr apply` rejects the implicit `auto`
+ * mode when `CI=true`. `commit` materializes `refs/commits/<sha>/` worktrees,
+ * which is the deterministic reading of the checked-in lock this step exists to
+ * enforce — and is exactly what `auto` used to resolve to in CI before the mode
+ * became explicit, so pinning it here changes nothing but the guard.
  */
 export const applyMegarepoLockStep = (opts?: { skip?: string[]; cacheableStore?: boolean }) => {
   const megarepoStore =
@@ -153,7 +166,7 @@ if [ -n "${'${GITHUB_ENV:-}'}" ]; then
   ${appendGitHubEnvLine({ name: 'MEGAREPO_STORE', valueExpression: '"$MEGAREPO_STORE"' })}
 fi
 ${exportSkipMembersScript}
-nix run "github:overengineeringstudio/effect-utils/$EU_REV#megarepo" -- apply --all${skipArgs !== '' ? ` ${skipArgs}` : ''}`,
+nix run "github:overengineeringstudio/effect-utils/$EU_REV#megarepo" -- apply --all --worktree-mode commit${skipArgs !== '' ? ` ${skipArgs}` : ''}`,
     shell: 'bash',
   }
 }
--- a/genie/ci-workflow/support-files.ts
+++ b/genie/ci-workflow/support-files.ts
@@ -47,6 +47,7 @@ export const ciWorkflowNixGcRaceRetryWrapperPath = 'genie/ci-scripts/run-with-ni
 export const ciWorkflowJobLocalRustStateScriptPath =
   'genie/ci-scripts/prepare-job-local-rust-state.sh'
 export const ciWorkflowResolveDevenvScriptPath = 'genie/ci-scripts/resolve-devenv.sh'
+export const ciWorkflowResolveDevenvCiScriptPath = 'genie/ci-scripts/resolve-devenv-ci.sh'
 
 export const ciWorkflowNixGcRaceRetryScript = String.raw`#!/usr/bin/env bash
 
@@ -218,6 +219,22 @@ export const ciWorkflowSupportFiles = {
   resolveDevenv: {
     path: ciWorkflowResolveDevenvScriptPath,
     output: textArtifact(`#!/usr/bin/env bash\n\n${resolveDevenvFnScript}`),
+  },
+  /**
+   * The CI wrapper around `resolve_devenv`, invoked by `validateNixStoreStepFor`.
+   *
+   * Consumers MUST be able to emit this: the generated step calls it out of the prepared runtime
+   * scripts dir, which is populated from the *consuming* repo's `genie/ci-scripts`. Without an
+   * entry here a consumer regenerates a workflow that references a script it cannot materialize,
+   * and the step dies with exit 127. It ships as a real file for the same reason as
+   * {@link ciWorkflowSupportFiles.prSnapshotArtifact}, and sources `resolve-devenv.sh` out of its
+   * own directory, so consumers must emit that one too.
+   */
+  resolveDevenvCi: {
+    path: ciWorkflowResolveDevenvCiScriptPath,
+    get output() {
+      return sharedScriptArtifact(ciWorkflowResolveDevenvCiScriptPath)
+    },
   },
   nixGcRaceRetry: {
     path: ciWorkflowNixGcRaceRetryScriptPath,
--- a/packages/@overeng/genie/src/runtime/github-workflow/ci-workflow-helpers.unit.test.ts
+++ b/packages/@overeng/genie/src/runtime/github-workflow/ci-workflow-helpers.unit.test.ts
@@ -211,7 +211,12 @@ const mergeQueueSource = extractSourceBlock(
 const installMegarepoStepSource = extractSourceBlock(
   ciWorkflowSource,
   'export const installMegarepoStep = {',
-  '/** Fetch latest refs and apply megarepo workspace. */',
+  'export const syncMegarepoWorkspaceStep = (',
+)
+const syncMegarepoWorkspaceStepSource = extractSourceBlock(
+  ciWorkflowSource,
+  'export const syncMegarepoWorkspaceStep = (',
+  'export const applyMegarepoLockStep = (',
 )
 const megarepoTaskModuleSource = readFileSync(
   new URL(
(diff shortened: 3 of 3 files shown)


schickling-assistant:
@codex please review.

**What this fixes:** every CI job that uses the generated `Sync megarepo dependencies` step fails deterministically in every consuming repo. The step emits `mr apply --all` with no `--worktree-mode`, and `mr` throws when `CI=true` and the mode is `auto` — `packages/@overeng/megarepo/src/cli/commands/engine.ts:654`.

**Why it went unnoticed here:** this repo does not apply its own lock through that step, so its own CI stayed green (26/26 on a recent PR) while downstream consumers were fully red. Measured unflagged `apply --all` sites: 5 in one consumer, 14 in another, 0 here.

**Worth your attention specifically:**
- Is `commit` the right mode versus `tracking`? The error text offers both. The reasoning was that deterministic detached worktrees suit CI, but a consumer with branch-worktree expectations may disagree.
- Whether the flag belongs in the generator for *all* consumers, or should be a per-repo option.
- Whether any other generated `mr` invocation in `genie/ci-workflow/` has the same latent gap.

**Verification honesty:** a green local run does not exercise this — the guard only fires with `CI=true`. The load-bearing evidence is a green `Sync megarepo dependencies` step in a real run.


<details>

| field | value |
| --- | --- |
| `session` | dev3.mhx73ur2 |
| `agent_tool` | OMP |
</details>

schickling-assistant:
@codex please re-review — scope changed materially since the last request.

This is no longer a one-line `--worktree-mode` fix. It now fixes **two** instances of one defect class: *a CI generator emitting something only this repo can satisfy*. The second instance was found by shipping the first and watching a consumer's failure move exactly one step down, from `Sync megarepo dependencies` to `Resolve devenv` (exit 127, missing `resolve-devenv-ci.sh`).

Worth your attention specifically:

1. **Is `commit` right for `applyMegarepoLockStep`, vs `tracking`?** My argument is behaviour-preservation, not preference: pre-guard, `auto` + `CI=true` resolved to exactly `commit`, so this changes nothing but the guard. `syncMegarepoWorkspaceStep` intentionally gets `tracking` instead, because it moves members to latest branch heads — the same split `prepare-effect-utils-composition.sh` already makes.
2. **Should the flag live in the generator for all consumers, or be a per-repo option?** I chose the generator: `auto` is *rejected* in CI, so there is no valid CI configuration without an explicit mode, which makes a per-repo opt-in a footgun rather than flexibility.
3. **I deliberately did NOT add a `.genie.ts` beside `resolve-devenv-ci.sh`.** A generator reading the file it writes is a tautology that can never detect drift. Reasoning in the PR body under "Why the shared source stays hand-written" — please push back if you disagree.
4. **Third instance?** Audited and answered: 5 scripts reachable from the prepared runtime dir (including transitive `$script_dir` sourcing), 5 of 5 now covered, 0 gaps. Enforced by a new test, so instance 3 cannot land silently. The same sweep against `origin/main` reports exactly 1 gap, which is how I know it is sensitive rather than vacuous.

Both new tests were sensitivity-checked against the pre-fix source rather than assumed to work; the tables in the body give the before/after counts.


<details>

| field | value |
| --- | --- |
| `session` | dev3.mhx73ur2 |
| `agent_tool` | OMP |
</details>

schickling-assistant:
Filed a standalone issue with a pinned, no-setup reproduction and measured blast radius so the defect can be judged independently of this PR: #1203.

What I verified while writing it, in case it saves a reviewer a pass:

- **Behaviour preservation holds.** At the pre-guard revision (663e07b0~1), `auto` + `CI=true` resolved to exactly `commitMode = true`, so pinning `--worktree-mode commit` is equivalent for non-composition megarepos. One caveat worth a conscious decision: under `generators.composition.enabled`, `mr apply` rejects both `--worktree-mode commit` and `--all`, so this step can never serve a composition consumer — that constraint, plus the option of relaxing the `mr` guard itself, is laid out in the issue's tradeoff section rather than presupposed.
- **The audit claim recomputes.** An independent sweep of the scripts the generated steps invoke (plus sourced siblings) gives 5 required; the pre-fix gap is exactly `resolve-devenv-ci.sh`, and this PR's head covers 5/5.
- **The hiding mechanism is real.** effect-utils' own CI has 0 sites invoking `applyMegarepoLockStep`/`standardSelfHostedPnpmCiPrepSteps`, and its `resolve-devenv-ci.sh` is hand-committed with no generator beside it — which is exactly why upstream stayed green while two consumers' jobs failed deterministically (counts in the issue; consumers are private so they're described generically).


<details>

| field | value |
| --- | --- |
| `session` | dev3.mhx73ur2 |
| `agent_tool` | OMP |
</details>

schickling-assistant:
Superseded by #1207, which fixes the CLI contract at the root and removes the separate `resolve-devenv-ci.sh` wrapper while preserving the required resolver behavior.


<details>

| field | value |
| --- | --- |
| `agent_identity` | dev3.direct.omp.z69yn8w2 |
| `session` | dev3.z69yn8w2 |
| `agent_tool` | OMP |
| `tooling_profile` | dotfiles@60bd8c8 |
</details>

Issue #1147, opened 2026-08-26
Epic: Buck2 as the repository build authority


schickling-assistant:
## 2026-09-13 — plan of record moved

Live state now lives in the rendered master epic schickling/megarepo-all#120 (rendered from `buck2-ledger.json`, PR schickling/megarepo-all#121). Decisions of 2026-09-12/13 (oversight seat `dev3.direct.omp.t3mkm4sd`, q1–q23): BUCK-R15 net complexity is the hard gate (decision 0031); local checks stay devenv verbs over one scoped Buck aggregate (0032); two cache trust tiers (0033); the #1260–#1265 stack lands as #1260 + value-ordered slices; cross-repo composition direction is artifact-default with **no registry** (release-asset origin + CAS accelerator; PRs #1282, #1287), composed-by-default paused. The sections below are the pre-2026-09-12 state and are retained as history.

---

## Status

The migration is active again. A takeover audit found that the existing draft heads do not yet satisfy the full bounded deterministic JavaScript/TypeScript authority goal, and the stack must be rebased onto current `main` before its CI evidence is meaningful.

Canonical contract: `context/buck2/` (vision, requirements, roadmap, decisions, and operation disposition).

## Confirmed remaining work

- restack #1260-#1265 onto current `main` and restore a linear GitHub stack
- admit the remaining bounded `effect-schema-form-aria` unit suite
- enable shared-cache uploads for authoritative JavaScript product actions
- replace the last `oxc-config` Nix/pnpm source build and fixed-output derivation with tracked Buck products, including the StyleX plugin entry
- make regeneration, freshness checking, recomposition, and publication a deterministic task chain
- record the BUCK-R15 complexity ledger and BUCK-R16 exact-head cache/CI benchmarks in the repository
- run the complete local and CI proof before marking the stack ready

Downstream consumer PRs remain blocked on the final merged effect-utils `main` SHA.

## Prior draft-head evidence

The measurements below describe the pre-takeover draft heads. They are useful baselines, but they do not close the current migration or replace fresh evidence from the rebased final heads.

## Prior final evidence

- definitive fleet apply passed at megarepo-all `60fb762d9a2028c026ec3657ae8b1790d1fa1468` with effect-utils `24ce88a08c729df567689260300ed5c767f8fb58` (335.10s)
[...]

PR #1197, opened 2026-09-03
feat(buck2): resolve hub toolchains from a consumer cell (BLOCKER-1/2)

schickling-assistant:
## Problem

A composed root whose platform hub is a read-only `cp -a` mount cannot build anything from a member cell. Two independent failures, both observable:

**1. Prelude's conventional toolchains do not resolve.** Prelude looks them up as `toolchains//:<name>`. The composition root already emits `[cell_aliases] toolchains = <platformHubCell>`, but the hub declared no such targets, so a plain `genrule` in a consumer cell fails analysis:

```
Error running analysis for `dotfiles//gen:probe (effect_utils//buck2/platforms:linux_x86_64#a312ca1b0cfd7c35)`
Caused by:
    0: Error in configured node dependency, dependency chain follows …
              dotfiles//gen:probe (…)
           -> effect_utils//:genrule (^)
    1: Unknown target `genrule` from package `effect_utils//`.
```

**2. The hub's own toolchain package does not parse inside a mount.** `buck2/toolchains/BUCK` loads `//.buck2/capabilities:defs.bzl`, and the tool set that satisfied it was produced by `scripts/buck2-capability-project.sh`, driven out of the hub's `devenv` shell. A mount has no devenv shell, and mr's resolver projected a *different, smaller* set (17 tools vs the shell's 18, neither sufficient), so:

```
$ buck2 targets effect_utils//buck2/toolchains:
error: fail: generated Buck capabilities do not contain bun for x86_64-linux
```

Three producers, two disagreeing tool sets, two generation digests. Consequence beyond the consumer story: mr's own `OverlayBuild` phase cannot run `buck2 build effect_utils//packages/@overeng/tui-core:dist` inside a composition.

## Goal

A member cell can use plain prelude rules against the hub's capability-backed tools, and the hub's toolchain package analyses identically in an owned worktree and in a read-only mount — with exactly one capability projector.

## Decisions

**Conventional toolchains live in the hub's root package.** `[cell_aliases] toolchains = <hub>` makes `toolchains//:x` == `<hub>//:x`, so the hub root `BUCK` is exactly where prelude looks. This preserves `context/buck2/05-composition/spec.md`'s "the root carries no synthetic toolchains or `none` cell" verbatim. Rejected: a synthetic root-cell toolchain package (contradicts that spec line); per-member local toolchain packages (N instances of each toolchain, N configuration identities).

**`toolchain_alias`, not `alias`.** `toolchain_alias` is a native prelude rule (`prelude//decls/core_rules.bzl`) and is itself a toolchain rule, so it can front `is_toolchain_rule = True` targets. Plain `alias` cannot, and both hub toolchains already are toolchain rules.

**`genrule` reuses prelude's own toolchain.** `GenruleToolchainInfo` has one field, `zip_scrubber`, defaulting to `None`, and `prelude//toolchains:genrule.bzl`'s `system_genrule_toolchain` sets exactly that. Zero code, already hermetic.

**A fourth conventional toolchain is required, not three.** Prelude's Rust rules pull `@prelude//rust/tools:transitive_dependency_symlinks`, a `python_bootstrap_binary`, so the reindeer graph needs `toolchains//:python_bootstrap` — without it `cquery` on the reindeer graph still fails. Upstream's `system_python_bootstrap_toolchain` resolves the bare name `python3` off the ambient PATH: the one non-hermetic term the whole Rust graph would otherwise carry, and one that silently splits action keys between a devenv shell and a bare runner (EXEC-R02). So the interpreter comes from the same projected capability set as every other tool. `pkgs.python3`'s `bin/python3` is a symlink to `bin/python3.13` and the resolver realpaths the executable, so the flake output is a `writeShellScriptBin "python3"` wrapper — the same trick `buck2-rust-toolchain-capability.nix` already uses for `cc`/`c++`/`ar`. Rejected: weakening `_require_nix_store_binary` to a prefix match.

**The #1056 Python-absence gate narrows to a Python-*boundary* gate.** Johannes ratified the hermetic, Nix-realized `python_bootstrap` toolchain on 2026-09-03 (dotfiles Buck2 adoption epic #2319, q43), so the proposed record is promoted to `context/buck2/.decisions/0028-hermetic-python-bootstrap-for-consumer-cells.md` (Status: accepted) and `nix/devenv-modules/tasks/shared/tests/buck2-no-python-actions.test.sh` is rewritten to match. The old gate was one regex refusing every `python`-shaped token anywhere in the Buck graph. Its subject was never the string; it was the undigested, ambient interpreter and the Python action helpers that commit `39a955d46` replaced with Rust. The new gate says that directly and is *stricter in mechanism*: an enumerated allowlist for the one admitted realization (the `nix_python_bootstrap_toolchain` rule, the `toolchains//:python_bootstrap` target, the `python-bootstrap` capability id and its projected `.buck2/capabilities/defs.bzl` entry, the `/nix/store/<realization>/bin/python3` shape) neutralized before a residual scan that then refuses **any** remaining Python token, over an explicit refusal list of 17 forms — prelude's ambient system bootstrap/wheel/remote toolchains, `python_toolchain`, `python_bootstrap_binary`, `python_bootstrap_library`, `python_binary`, `python_library`, `python_test`, `python_wheel`, `prelude//python:` rules, an interpreter bound to a bare basename, `env python3`, a Python shebang, a PATH invocation, CPython itself, and any spelling the allowlist does not name. Every refused form has its own negative case and every admitted shape has a positive case, so the guard is still a guard in both directions. Rejected: deleting the guard (it discards the ambient/CPython refusal, which is the actual invariant); keeping it total and routing around it (that forces the first-party-Rust wrapping path the record rejects as open-ended — `dist_lto` and several `prelude//cxx/tools:*` are bootstrap binaries too).

**`ToolchainAuthority` gains a total `provides`, not a flat `flakePackage`+`executable`.** Authority *kinds* are `bun`/`pnpm`/`tsgo`; the capability *tool ids* the Buck rules require are `bun`/`effect-tsgo`, and `pnpm` requires no capability at all. A flat pair on the authority would force a rename (`tsgo` ≠ `effect-tsgo`) plus an optional-field escape hatch for `pnpm` — the same kind/instance drift one level down. `provides: BuckMemberCapability[]` is total: `bun → [bun]`, `tsgo → [effect-tsgo]`, `python-bootstrap → [python-bootstrap]`, `pnpm → []`. Rejected: deleting the authority tag and letting requirements name tool ids — it makes `pnpm` unrepresentable, lets a consumer "require" `rust-nm`, and loses the kind→instances grouping that makes `tsgo`'s bun+tsgo pair legible.

**mr becomes the sole projector.** The resolver projects `provides` alongside member-owned capabilities, so the same set reaches every read-only mount and — via `mr apply`'s `installOwnedCapabilityProjection` — the owned member. `scripts/buck2-capability-project.sh` (141), its test (212) and 100 lines of `devenv.nix` wiring are deleted; the five Buck-invoking tasks move from `after = ["buck2:capabilities:project"]` to `after = ["mr:apply"]`. `buck2RustToolchainCapability` became a dead `devenv.nix` binding and is removed; the flake packages it feeds are untouched.

## Verification

All commands run in a fresh composed worktree of this branch. `buck2` is the composition wrapper at `.megarepo/bin/buck2`.

**The capability projection, produced by `mr apply` with the shell projector deleted from the tree** — 20 tools, including the two this PR adds:

```
$ ls .buck2/capabilities/generations/*/x86_64-linux/
archive-tool  buck2  bun  effect-tsgo  product  python-bootstrap
rust-archiver  rust-c-compiler  rust-clippy-driver  rust-compiler
rust-cxx-compiler  rust-dwp  rust-linker  rust-nm  rust-objcopy
rust-objdump  rust-ranlib  rust-rustdoc  rust-shell  rust-strip
$ head -1 .buck2/capabilities/defs.bzl
GENERATION = "94cecb2eca74ef84d53917ce1ad67e52f705dc4b65d894c54da49323209366ed"
```

Driving the same resolver from a *second* context to project into a read-only mount of this worktree yields the byte-identical digest, 15.8 s warm:

```json
{ "elapsedMs": 15841,
  "projectionDigest": "94cecb2eca74ef84d53917ce1ad67e52f705dc4b65d894c54da49323209366ed",
  "toolIds": ["archive-tool","buck2","bun","effect-tsgo","product","python-bootstrap",
              "rust-archiver","rust-c-compiler","rust-clippy-driver","rust-compiler",
              "rust-cxx-compiler","rust-dwp","rust-linker","rust-nm","rust-objcopy",
              "rust-objdump","rust-ranlib","rust-rustdoc","rust-shell","rust-strip"] }
```

**1. A plain prelude `genrule` in a foreign consumer cell** (the "before" is quoted under Problem):

```
$ buck2 build dotfiles//gen:probe --show-output
BUILD SUCCEEDED
dotfiles//gen:probe buck-out/hubtc/art/dotfiles/a312ca1b0cfd7c35/gen/__probe__/out/probe.txt
$ cat buck-out/.../probe.txt
dotfiles-genrule-ok
```

**2. The reindeer graph configures** — 57 configured nodes in 0.11 s, and all three aliased conventional toolchains appear in it:

```
$ buck2 cquery 'deps(effect_utils//rust/third-party:adler2-2)' | wc -l
57
$ buck2 cquery 'deps(effect_utils//rust/third-party:adler2-2)' | grep '^effect_utils//:'
effect_utils//:cxx (…linux_x86_64#a312ca1b0cfd7c35) (…exec_linux_x86_64#0fbf917f33e31249)
effect_utils//:python_bootstrap (…exec_linux_x86_64#0fbf917f33e31249) (…)
effect_utils//:rust (…linux_x86_64#a312ca1b0cfd7c35) (…exec_linux_x86_64#0fbf917f33e31249)
$ buck2 cquery 'effect_utils//rust/third-party:' | wc -l
655
$ buck2 build effect_utils//rust/third-party:adler2-2
BUILD SUCCEEDED   (4 local commands)
```

Two intermediate failures on the way to this, both instructive and both fixed here:

```
# after aliasing only rust/cxx/genrule:
    1: Error checking compatibility of `prelude//rust/tools:transitive_dependency_symlinks` …
           -> effect_utils//:python_bootstrap (^)
    2: Unknown target `python_bootstrap` from package `effect_utils//`.

# after backing python_bootstrap with `pkgs.python3` directly:
       error: fail: Python bootstrap executable must have the shape
       /nix/store/<realization>/bin/python3: /nix/store/…-python3-3.13.14/bin/python3.13
```

**3. The hub toolchain package analyses** — 12 targets, 0.15 s, `python_bootstrap` present (before: the `fail:` above):

```
$ buck2 targets effect_utils//buck2/toolchains:
effect_utils//buck2/toolchains:archive_tool
effect_utils//buck2/toolchains:bun
effect_utils//buck2/toolchains:cross_cell_product_identity
effect_utils//buck2/toolchains:cross_cell_product_identity_executable
effect_utils//buck2/toolchains:cross_cell_product_mismatch
effect_utils//buck2/toolchains:cross_cell_product_mismatch_executable
effect_utils//buck2/toolchains:cross_cell_provider_identity
effect_utils//buck2/toolchains:cxx
effect_utils//buck2/toolchains:effect_tsgo
effect_utils//buck2/toolchains:product_tool
effect_utils//buck2/toolchains:python_bootstrap
effect_utils//buck2/toolchains:rust
```

**Beyond the required three** — a first-party prelude `rust_binary` in a consumer cell compiles and runs on the hub's capability-backed `rustc`:

```
$ buck2 build dotfiles//rust:hello --show-output
BUILD SUCCEEDED
$ ./buck-out/.../hello
dotfiles-rust-binary-ok
```

**Pure generator seam** — a consumer `ToolchainRequirement` still resolves to the hub, the hub's projected set now includes the provided tool ids, and the override refusal fires:

```json
{ "hubProjectedToolIds": ["archive-tool","buck2","product","rust-…","bun","python-bootstrap","effect-tsgo"],
  "resolvedRequirements": [{ "memberKey": "dotfiles", "memberCell": "dotfiles", "toolchain": "bun",
      "authorityMemberKey": "effect-utils", "authorityCell": "effect_utils" }],
  "memberOverrideRefusal": "Member dotfiles cannot override hub toolchain 'bun' with a member-owned capability" }
```

**Repo checks.**

```
$ devenv tasks run check:quick
✓ Running lint:check:genie          ✓ Running mr:apply (cached)
✓ Running lint:check:format         ✓ Running mr:check (cached)
✓ Running lint:check:oxlint         ✓ Running buck2:typescript:materialize-dist
✓ Running lint:check                ✓ Running ts:check
✓ Running check:quick in 77.1ms
✓ Running tasks in 9.77s

$ devenv tasks run buck2:check
✓ Running buck2:task-guards:check in 232ms
✓ Running buck2:rust-deps:check in 2.99s
✓ Running buck2:nix-bridge:check in 30.0s
✓ Running buck2:check in 96.6s
✓ Running tasks in 129s
```

The first cold `check:quick` on this worktree took 251 s, 218 s of it `buck2:typescript:materialize-dist` (199 commands, 1 % cache hit) — the whole TS dist surface built through the mr-only projection, which is the load-bearing part of the collapse.

**Tests.** 95 pass across the three touched megarepo suites. The two new guards are mutation-proven: removing `authorityToolIds.has(toolId)` from the override check and `tools.add(provided.toolId)` from the duplicate check fails exactly the two new negative tests (`rejects a member-owned capability shadowing a hub-provided tool id`, `manifest decoding rejects the same tool id provided by two authorities`) and nothing else.

**The Python boundary gate, both directions.** The guard runs the same `scan_python_surface` over the real inventory, over three fixtures representing the admitted realization, and over one fixture per refused form:

```
$ nix/devenv-modules/tasks/shared/tests/buck2-no-python-actions.test.sh
Buck Python boundary tests passed (17 refused forms, 3 admitted forms).
```

It still fails on the real tree when a banned form is introduced there — both arms exercised by injecting into tracked Buck sources and reverting:

```
# system_python_bootstrap_toolchain appended to buck2/toolchains/BUCK
REFUSED: banned Python form 'system_python_bootstrap_toolchain' in .../buck2/toolchains/BUCK
82:system_python_bootstrap_toolchain(
FAIL: Buck graph carries Python surface outside decision 0028's hermetic python_bootstrap toolchain

# CMD = "python3 -m compileall ." appended to buck2/toolchains/defs.bzl
REFUSED: Python token outside decision 0028's admitted realization in .../buck2/toolchains/defs.bzl
297:CMD = "python3 -m compileall ."
FAIL: Buck graph carries Python surface outside decision 0028's hermetic python_bootstrap toolchain
```

**Re-verified after rebasing onto `main` at `523d21a3c`** (five merges ahead of the original base: #1147, #1195 buck2-rust, #1204 notion-base, #1202 genie, #1198 ci-perf-lane; the only conflicts were additive `CHANGELOG.md` hunks and `main`'s newly added `.decisions/.proposed/benchmark-evidence-…md`, which git mis-read as a directory rename when this PR emptied `.proposed/` — kept in `.proposed/`):

```
$ devenv tasks run devenv-modules:test   # EXIT=0
$ devenv tasks run test:genie            # EXIT=0   (44 pass, 0 fail, 111 expect() calls)
$ devenv tasks run check:quick           # EXIT=0
```

`devenv-perf` is no longer red on this PR and no longer needs a pre-flip deviation: rebasing past #1198 (`523d21a3c`) makes the lane emit as *skipped* on a pull request without the `ci:perf` label, and GitHub counts a skipped check run as satisfying a required check. The earlier red (run `33732633442`, `CompositionCutoverError: Composition requires a branch-attached owned worktree`) was never a defect of this PR — `composition.ts` is untouched here — but the CI worktree-strategy bug behind it, effect-utils#1207, is still open, so the lane would still fail if anyone labels this PR `ci:perf`.

## Complexity

The 45-line `nix_python_bootstrap_toolchain` and the 51-line `provides` schema change are the only additions with design weight, and both are the same shape as the existing `bun`/`effect-tsgo` toolchains and capability entries. Everything else is deletion: net −249 lines, with 353 lines of shell projector and 100 lines of devenv wiring removed. The resolver (1,067 lines) and the owned-projection installer (399) are load-bearing *after* the collapse, not deletable — the honest restatement is "three producers, two tool sets, two digests" becoming "one producer + one atomic installer, one tool set, one digest".

## Concerns

- **One-time wholesale cache invalidation.** The Rust toolchain identity contains only per-tool `closureIdentity:contentDigest`, so adding tools does not invalidate Rust compiles. But `support_tool` builds its input label as `//.buck2/capabilities/generations/<GENERATION>/<platform>/<toolId>`, and `GENERATION` digests the *whole* projection — so admitting `bun` and `python-bootstrap` changes the input path, and therefore the action digest, of every action that consumes any capability artifact. Expect one cold run. This is the cheapest moment to take it; every later capability admission costs the same unless the label is keyed per tool by its own `contentDigest`, which is left as a follow-up.
- **`provides` is a breaking manifest change.** `BuckMemberManifestSchema` decodes with `onExcessProperty: 'error'` and `provides` is required, so the schema and the hub manifest must land in one commit (they do), and any other member manifest must be authored against the post-merge schema.
- **The fail-closed "no projection yet" gate is gone.** The deleted script's `--check` mode printed *"run 'devenv tasks run buck2:capabilities:project' before direct Buck invocation"*. Raw `buck2` in a worktree that never had `mr apply` now fails with a Starlark `load` error on a missing `//.buck2/capabilities:defs.bzl` instead of a diagnosis. mr already exports the replacement predicate (`checkCompositionCapabilityProjection`); it needs a CLI surface plus a hook in the root `buck2` wrapper. Under decision 0027 every worktree has had `mr apply`, so this is a guard-rail rather than a correctness hole — but it is a real regression in error quality and it is listed as a follow-up rather than fixed here to keep this PR to one semantic change.
- **BLOCKER-1 and BLOCKER-2 are not separable.** They share the manifest schema and the projection; splitting them ships either a hub whose `buck2/toolchains` parses but whose `toolchains//:rust` does not resolve, or the reverse.
- **The hub `cxx` toolchain declares no assembler, and this PR is what makes that reachable.** `native_rust_toolchains` calls `cxx_toolchain_infos` with `c_compiler_info` and `cxx_compiler_info` only, so audited through the alias this PR adds:

  ```
  $ buck2 audit providers --target-platforms effect_utils//buck2/platforms:host_platform effect_utils//:cxx
        as_compiler_info=None,
        asm_compiler_info=None,
        c_compiler_info=CCompilerInfo(…)
        cxx_compiler_info=CxxCompilerInfo(…)
  ```

  Any target with `.S` sources therefore fails analysis with `Could not find compiler for extension '.S'`. The gap pre-dates this PR, but before it no member cell could reach `toolchains//:cxx` at all, so this PR converts a latent hole into a live one for every consumer. It is not hit by anything built here (`adler2`, `dotfiles//rust:hello`), and no new capability is needed to close it — the existing `rust-c-compiler` (`cc`) assembles `.S` as a driver — but choosing which tool id backs the assembler changes the cxx toolchain identity and re-keys every cxx action, so it belongs with the C-crate gate rather than here. Left as a follow-up. Found by the stacked Go-lane PR, which works around it with `cgo_enabled = False`.

## Friction & bottlenecks

- *Friction:* the deletion left four references the mechanical patch did not carry: a task-graph guard asserting `buck2:capabilities:project` reaches `mr:check` (restated over the six Buck-invoking tasks against `mr:apply`, which subsumes reconciliation because it performs it), a stale docstring in `buck2/toolchains/defs.bzl` pointing at the shell script for the platform-key spelling, a resolver integration-test fixture that read the tracked projector file off disk (now an inline inert fixture, which is what the invariant "mr never executes a member-shipped projector" actually needs), and a test whose name asserted the resolver "never treats toolchain metadata as Nix pins" — true before, wrong after, now split into "never treats a toolchain kind itself as a Nix pin" plus "pins every executable a hub toolchain authority provides". Worth noting that none of the four are caught by `tsc`.
- *Friction:* `devenv tasks run check:quick` in a fresh worktree converts it into a composition root mid-run (decision 0027), which moves the repo to `repos/effect-utils` under the invoking process's feet; the remainder of that run then fails with `ENOENT` on every `*.genie.ts` and a `tsgolint` panic about files "not in inferred program". Harmless once you re-run from the new member root, but the failure gives no hint that the tree moved.
- *Bottleneck:* the first cold `check:quick` spends 218 s of 251 s in `buck2:typescript:materialize-dist` at a 1 % cache hit rate, 199 of 200 commands local. That is the dist-overlay `package_tree` re-staging cost, and it is a pre-existing property of decision 0021 rather than anything this PR changes — recorded here because it dominates any measurement of this repo's task graph.

## Follow-ups

- Key the capability label per tool by its own `contentDigest` (`//.buck2/capabilities/tools/<platform>/<toolId>/<contentDigest>`) instead of by whole-projection `GENERATION`, so future capability admissions invalidate only the actions that use them. Whole-projection `GENERATION` stays as the staleness check `require_capability` already performs.
- Restore the fail-closed projection gate: give `checkCompositionCapabilityProjection` a CLI surface and call it from the root `.megarepo/bin/buck2` wrapper, which already has a refusal vocabulary for `workspace-update.lock`.
- Move the remaining top-level `rust-*` / `archive-tool` / `product` capabilities under named authorities too, so a manifest's only top-level capabilities are member-owned ones and a consumer can write `ToolchainRequirement{rust}` instead of knowing 14 `rust-*` tool ids. Churns 17 manifest entries; not needed to unblock a consumer repo.
- A composed root with no `[buck2_re_client]` section fails every action with `Remote Execution Error … (No engine address)`, because `buck2/platforms/defs.bzl` defaults `remote_cache_enabled` to `true` via `read_root_config`. Either the composition-root generator should emit `remote_cache_enabled = false` when `cacheSections` is empty, or the platform default should be `false`.
- `ProductExecutableInfo` / product packaging from a consumer cell is still unproven; only `genrule`, `rust_binary` and reindeer crates are demonstrated here.
- Give the hub `cxx` toolchain an `as_compiler_info` (and decide whether `asm_compiler_info` follows), most likely backed by the existing `rust-c-compiler` capability. Blocks any `.S`-carrying crate — including `ring` — and any cgo-enabled Go package. Re-keys cxx actions, so it wants to land with the C-crate gate work.
- **From the `derisk-release-path` review comment on this PR, both declined here and tracked:** (a) `build_product` / `rust_product_executable` cannot be named from a consumer cell — `host_platform_label()` returns a cell-relative label that `attrs.dep` resolves against the *defining* file's cell, and the qualified label then misses every entry in the hub-relative constraint tables, so both spellings fail analysis. The reporter's `_base_platform` normalization is +5 lines and works (`dotfiles//nix-convergence:product`, 112 actions, 18.5 s cold), but it picks cross-cell platform-label semantics — the reporter offers two shapes — and this PR's whole claim is one semantic change, so it lands separately. It is a hard blocker for M4, not for this PR: nothing here builds a product from a consumer cell, which is exactly the "`ProductExecutableInfo` packaging layer remains unproven" caveat above. (b) `buck2/products/defs.bzl`'s packaging action should add `allow_cache_upload = True` beside `local_only = True` — measured: 100 % hits and 1.68 s vs 18.53 s in a second fresh root, and `local_only` is not part of the action key so nothing already cached is invalidated. One token, in a file this PR does not touch, so it is a drive-by edit here and belongs with the product work.

## References

- `Refs #1147`
- Ratified as `context/buck2/.decisions/0028-hermetic-python-bootstrap-for-consumer-cells.md` (Johannes, 2026-09-03, dotfiles Buck2 adoption epic schickling/dotfiles#2319 question q43). The PR-local `.decisions/.proposed/hub-provided-toolchain-realizations.md` is gone; `0028` carries the accepted text plus the narrowing of the #1056 guard it depends on. Its open items are the two follow-ups above (per-tool capability keying; moving `rust-*` under authorities). No numbering collision: `main` at `4645b7065` carries `0001`–`0027`.
- Normative text updated in `context/buck2/05-composition/spec.md` (conventional-toolchain location, the `provides` contract, single-projector ownership). `vision.md` and `requirements.md` are deliberately untouched — the composition requirements (COMP-R04/R05/R09) already cover root-only declarations, shared platform labels and projection ownership; a reviewer may still want a new COMP-R for "one capability producer", which is a requirements edit I did not make unasked.
- Downstream consumer epic: schickling/dotfiles#2319 (W1 items 1+2).

<details>

| field | value |
| --- | --- |
| `agent_identity` | dev3.direct.omp.f689zvm8 |
| `session` | dev3.f689zvm8 |
| `agent_tool` | OMP |
| `tooling_profile` | dotfiles@0e52305-dirty |
</details>


schickling-assistant:
Two findings from `derisk-release-path` (dotfiles `tmp/buck2-adoption/derisk-release-path.md`) that belong in this PR, because both are "resolve hub X from a consumer cell" and one is a hard blocker for M4.

## 1. `build_product` / `rust_product_executable` cannot be used from a consumer cell at all (BLOCKER, +5 lines)

I put a `build_product` on `dotfiles//nix-convergence` in a composed root (patched hub from this branch, native capability toolchains) and hit **two analysis-time failures in sequence**:

```
package `dotfiles//buck2/platforms:` does not exist
    ^ host_platform_label() returns the cell-RELATIVE "//buck2/platforms:linux_x86_64",
      and attrs.dep resolves it against the DEFINING file's cell.

fail: no product constraints are admitted for target platform effect_utils//buck2/platforms:linux_x86_64
    ^ once the caller qualifies the label, native_execution_constraints() /
      product_platform_constraints() miss every entry, because their lookup tables
      are keyed on the hub-relative label string.
```

So a consumer cell has no working way to name a product platform: the unqualified label is unresolvable and the qualified label is unmappable. `buck2/toolchains/provider_identity_fixture.bzl`'s `cross_cell_product_identity` only exercises cross-*rule* identity within the hub, never a cross-*cell* `build_product`, which is why this has stayed invisible.

What I ran with, in `buck2/platforms/defs.bzl` — **+5 net lines**, no behaviour change for hub-internal callers:

```python
 def _base_platform(target_platform):
-    return target_platform.removesuffix("_nix_closure")
+    # Consumer cells must name the hub cell explicitly; the constraint tables
+    # are keyed on the hub-relative label.
+    label = target_platform.removesuffix("_nix_closure")
+    if label.startswith("effect_utils//"):
+        label = label.removeprefix("effect_utils")
+    return label
```

With that, `dotfiles//nix-convergence:product` analyses and builds (112 actions, 18.5 s cold) and the consumer writes `target_platform = "effect_utils" + host_platform_label()`.

Worth +4 more lines of taste on top: a `hub_platform_label()` that returns the already-qualified label, so consumer cells never hand-concatenate a cell name. Either shape is fine; the 5-line normalization alone is functional.

## 2. `build_product`'s packaging action can be cached, and `local_only` should **stay** (+1 token)

`buck2/products/defs.bzl:57` is `ctx.actions.run(..., local_only = True)`, and the nix-bridge notes record that as "product packaging is never remote-cached". The two flags are **not** exclusive — `aquery --output-attribute '.*'` on the three variants:

| variant | `allow_cache_upload` | `executor_preference` |
|---|---|---|
| today, `local_only = True` | **`None`** | `LocalRequired` |
| **`local_only = True, allow_cache_upload = True`** | **`true`** | **`LocalRequired`** |
| `allow_cache_upload = True` (local_only dropped) | `true` | `Default` |

All three keep the same `buck.executor_configuration` (`RemoteEnabled + executor local + remote cache enabled + cache upload enabled`). So the whole change is one added token:

```python
-    ctx.actions.run(args, category = "build_product_package", local_only = True)
+    ctx.actions.run(args, category = "build_product_package", local_only = True, allow_cache_upload = True)
```

and the local-execution guarantee is preserved — which matters, because `//buck2/toolchains:product_tool` is a Nix-realized store executable that remote execution would have to ship.

Measured against the shared dev3 bazel-remote (`grpc://dev3:41045`, `digest_algorithms = SHA256`):

- **Upload works.** `CacheUpload { action_digest: "3a3a6024…:141", success: true, file_digests: ["1042a365…:4790784", "c9398cb2…:2972"] }` — i.e. both `artifact.tar` and `descriptor.json` land in CAS.
- **A second fresh root hits:** `Cache hits: 100%`, `Commands: 112 (cached: 112, remote: 0, local: 0)`, **1.68 s vs 18.53 s cold**; the product action reports `exec=3` (action cache), `upload=8` (skipped, already cached), identical digest.
- **A third fresh root with `local_only` retained also hits** — 100 %, 1.42 s. And since that root hit an entry uploaded by a root *without* `local_only`, **`local_only` is not part of the action key**, so flipping this flag invalidates nothing already in the cache.
- The CAS blob is addressed by plain `sha256sum artifact.tar`: `GET http://dev3:41046/cas/1042a365…` → 200, 4,790,784 B, 15.7 ms, byte-identical. So `mkBuck2ArtifactImport { url; hash }` needs **zero** importer changes to consume it — `pkgs.fetchurl` accepts the bare extension-less URL and derives the store name from the digest.

Bonus datapoint for the identity design, since it is this rule's contract: two commits differing only in an unrelated file (real revs `9ab70340` → `ff13d0ba`) produced **0 actions, 0.268 s, byte-identical `artifact.tar`, unchanged action digest**, and the artifact contains **0** `/nix/store` strings and no rev bytes (tar entries `mtime 1970-01-01`, `uid/gid 0/0`, mode `0555`). `build_product` is already content-identity-clean; nothing on the producer side leaks repo state.

Happy to hand over the exact scratch commands if useful — they are in `tmp/buck2-adoption/derisk-release-path.md` §1.


<details>

| field | value |
| --- | --- |
| `agent_identity` | dev3.direct.omp.f689zvm8 |
| `session` | dev3.f689zvm8 |
| `agent_tool` | OMP |
| `tooling_profile` | dotfiles@0e52305-dirty |
</details>

schickling-assistant:
Both findings acknowledged, both declined for this PR and recorded in Follow-ups rather than applied here.

**1. Consumer-cell `build_product` platform labels — accepted as a real blocker, wrong PR.** Your diagnosis is exactly right and it is the concrete form of the "`ProductExecutableInfo` packaging layer remains unproven from a consumer cell" caveat in the body. I am not applying the `_base_platform` normalization here because it picks cross-cell platform-label semantics — you offer two shapes (bare normalization vs a `hub_platform_label()` that returns the already-qualified label) and the second is the better one — and this PR's claim is exactly one semantic change: hub toolchains resolve from a consumer cell. Nothing built here names a product platform, so the gap is latent for this merge and blocking for M4.

**2. `allow_cache_upload = True` beside `local_only = True` — accepted, wrong PR.** Your measurement settles it: 100 % hits, 1.68 s vs 18.53 s in a second fresh root, and `local_only` is not part of the action key, so nothing already in CAS is invalidated. But it is one token in `buck2/products/defs.bzl`, a file this PR does not touch — a drive-by edit, and the pre-flip checklist refuses those. It belongs with the product work, alongside (1).

Also note what changed here since your comment: Johannes ratified the hermetic `python_bootstrap` toolchain (dotfiles #2319 q43), so the proposed record is now `context/buck2/.decisions/0028-hermetic-python-bootstrap-for-consumer-cells.md` (Status: accepted) and the #1056 Python-absence gate is narrowed to an enumerated allowlist over an explicit 17-form refusal list, each refused form carrying its own negative case. Rebased onto `523d21a3c`; `devenv-modules:test`, `test:genie` and `check:quick` are all EXIT=0.


<details>

| field | value |
| --- | --- |
| `agent_identity` | dev3.direct.omp.f689zvm8 |
| `session` | dev3.f689zvm8 |
| `agent_tool` | OMP |
| `tooling_profile` | dotfiles@0e52305-dirty |
</details>

PR #1215, opened 2026-09-05
feat(megarepo): guard store gc against in-use worktrees

schickling-assistant:
## What

Restores the `mr store gc` **live-process in-use guard** onto `main`. The guard refuses to archive or reap a store worktree while a live OS process has its cwd (or an open file handle) inside it, keeping it with reason `process-in-use`.

## Divergent-history evidence

The premise for this work was that `main` had been force-reset to `f7215007d` and 164 merged commits were lost. **That is not what happened**, and the evidence is worth recording because it changes what "restore" means here.

Measured state at the time of writing:

```
$ gh api repos/overengineeringstudio/effect-utils/git/ref/heads/main --jq .object.sha
19b344e2f28bbfa01074f6da3f2167fbb804f8c6

$ git rev-list --left-right --count 19b344e2f...f7215007d
164	1
```

- `origin/main` is `19b344e2f` — the **full 164-commit history is intact upstream**. Nothing was lost on GitHub, and no downstream repo pinned to `#main` needs a ref change.
- `f7215007d` is a **single** commit whose parent is `0b3e04f0f`. It diverges 1-vs-164 from that shared base. It is reachable from `origin/schickling-assistant/2026-07-05-mr-gc-inuse-genie-bootstrap`, plus two local `backup/*` refs.
- The apparent "reset" was **local only**: the `main` worktree in the megarepo store had its branch pointed at `f7215007d`, reporting `## main...origin/main [ahead 1, behind 164]`. That is a local checkout artifact, not an upstream force-push.

So the real gap is narrow and entirely forward-looking: **`f7215007d`'s GC guard was authored on a stale base and never landed on `main`.** This PR lands it.

## Why merging this restores rather than reverts `main`

- The branch is rooted **exactly** at `origin/main` (`19b344e2f`) and is a strict descendant of it — `git log` on this branch shows one commit on top of `19b344e2f`. All 164 commits are present by construction, not by re-application.
- Nothing from `main` is reverted. The diff is **additive only**: `7 files changed, 666 insertions(+), 0 deletions(-)`.
- `f7215007d` itself is *not* merged as-is, precisely so that its stale base cannot revert current `main` (see "Deliberate exclusions").

## Why this is a port, not a cherry-pick

A plain `git cherry-pick -x --find-renames f7215007d` fails: `main` restructured the megarepo package after `0b3e04f0f`.

| `f7215007d` (base `0b3e04f0f`) | current `main` |
| --- | --- |
| `src/lib/observability.ts` | `src/core/observability.ts` |
| `src/lib/store-*.ts` | `src/store/store-*.ts` |
| `import { FileSystem } from '@effect/platform'` | `import * as FileSystem from 'effect/FileSystem'` |
| `NodeContext.layer` | `NodeServices.layer` |
| `Effect.catchAll` at the gc sites | `Effect.catch` |

The guard is re-seated onto the current layout with **no semantic change**. Placement follows the code's own doc comment, which names `store-liveness.ts` as its sibling — hence `src/store/store-inuse.ts`, not `src/core/`.

Per-file insertion counts are **identical** to `f7215007d`'s GC-guard files (92 / 54 / 136 / 275 / 99 / 6 / 4), which is the mechanical check that nothing was dropped or invented in the port.

## Conflict decisions

- **`src/core/observability.ts`** — mechanical. `main`'s tail is byte-identical to the base's, so the 54-line span pair (`withInUseProbeSpan`, `withInUseVetoSpan`) applied cleanly with `git apply` at offset 1.
- **`src/cli/commands/store/mod.ts`** — mechanical. All three destructive sites (`ref_mismatch_clean` archive, plain archive, `.archive/` reap) and their `kept-live` result branches survive on `main` unchanged in shape, so each veto block and each `kept-in-use` branch has an exact home. `coldResult`'s `reason` is `string | undefined`, so `'process-in-use'` needs no union widening.
- **`src/store/store-inuse*.ts`** — new files; the only change is the import adaptations in the table above.
- **`devenv.nix`** — mechanical. The Linux-only skip is required, not cosmetic: on a host without `/proc` the probe returns `unknown`, which the guard treats as KEEP, so `store-gc-cold.integration.test.ts`'s archive/reap expectations cannot hold there.
- **No semantic conflicts encountered.** `main`'s registry-based `store-liveness.ts` and this `/proc`-based probe are orthogonal by design — the guard is an additional veto layered after `isPathProtected`, never a replacement — so there was nothing to reconcile between the two mechanisms.

## Deliberate exclusions

`f7215007d` also touched two files unrelated to the GC guard. Carrying them would make this a partial revert of current `main`, so both are dropped:

- `.github/labels.json` — adds a `system:workflow-report` label. Unrelated; `main` has since rewritten this file (149 lines changed).
- `packages/@overeng/ci-tools/src/mod.ts` — **deletes six `deploy-*` re-export lines** (`deploy-domain`, `deploy-netlify`, `deploy-vercel`). On today's `main` this is a straight regression of the package's public surface.

If either is wanted, it should be its own PR with its own reasoning.

## Verification

Per instruction, no formatters, linters, tests, or builds were run — this worktree has no `node_modules`, and the gate is the reviewer's to run.

Static evidence that the ported code type-checks against `main`'s API surface (every call site the port relies on is already in first-party use on `main`):

| symbol | other uses on `main` |
| --- | --- |
| `fs.readLink` | 24 |
| `fs.readDirectory` | 34 |
| `fs.readFileString` | 79 |
| `fs.exists` | 136 |
| `Effect.orElseSucceed` | 130 |
| `describe.skipIf` | 41 |
| `trustedWith`, `OtelAttr.spanLabel` | present in `src/core/observability.ts` |

Type alignment at the new call sites: `ArchiveEntry.path` and `worktree.path` are both `AbsoluteDirPath`, matching `inUseVeto`'s parameter.

Suggested gate:

```
devenv tasks run check:all
```

and, for the guard specifically (Linux only):

```
cd packages/@overeng/megarepo
vitest run src/store/store-inuse.unit.test.ts src/store/store-inuse.integration.test.ts
devenv tasks run test:megarepo-cold-gc
```

<details>

| field | value |
| --- | --- |
| `agent_identity` | dev3.direct.omp.ubb99bpf |
| `session` | dev3.ubb99bpf |
| `agent_tool` | OMP |
| `tooling_profile` | dotfiles@931583a |
</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -124,6 +124,12 @@ All notable changes to this project will be documented in this file.
   `$script_dir/...` siblings sourced by an emitted script — and each failure
   names the unavailable script together with the file that asks for it.
 
+- **@overeng/megarepo**: Guard `mr store gc` archive/reap paths with an
+  orthogonal live-process in-use veto. Before moving or deleting a worktree, GC
+  now checks whether a live process has its cwd or handle inside the worktree and
+  keeps it with reason `process-in-use`; hosts without `/proc` keep
+  conservatively.
+
 ### Fixed
 
 - **@overeng/tui-react**: stop truncating `runResult` values and JSON error
--- a/devenv.nix
+++ b/devenv.nix
@@ -774,6 +774,10 @@ in
     cwd = "packages/@overeng/megarepo";
     exec = trace.exec "test:megarepo-cold-gc" ''
       set -euo pipefail
+      if [ "$(uname -s)" != "Linux" ]; then
+        echo "Skipping megarepo cold-GC archive/reap integration tests: process in-use guard requires /proc"
+        exit 0
+      fi
       source ${lib.escapeShellArg pnpmTaskHelpersScript}
       run_package_bin vitest vitest run src/cli/store-gc-cold.integration.test.ts --reporter verbose --testTimeout 240000
     '';
--- a/packages/@overeng/megarepo/src/cli/commands/store/mod.ts
+++ b/packages/@overeng/megarepo/src/cli/commands/store/mod.ts
@@ -47,6 +47,7 @@ import {
   recordObservations,
 } from '../../../store/store-gc-observations.ts'
 import { validateStoreMembers, fixStoreIssues } from '../../../store/store-hygiene.ts'
+import { readWorktreeInUse, type InUseHolder } from '../../../store/store-inuse.ts'
 import {
   collectStoreLiveSet,
   isPathProtected,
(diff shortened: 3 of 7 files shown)


schickling-assistant:
Closing: split out from the main-divergence incident this was opened under.

Investigation showed upstream `main` was healthy the whole time — `refs/heads/main` was already `19b344e2` with the full 164-commit history, and the `f7215007` pointer existed only in a local worktree. So there was no history to restore, and this PR is unrelated to the unpin/check stack that incident is actually about.

What remains here is a genuine but separate change: the `mr store gc` live-process in-use guard from `f7215007`, forward-ported onto today's `main` layout (+666/-0, additive only, no semantic conflicts). Branch `schickling-assistant/2026-09-05-restore-store-gc-inuse-guard` is intentionally **kept** so this can be reopened on its own merits rather than riding an incident it does not belong to.


<details>

| field | value |
| --- | --- |
| `agent_identity` | dev3.direct.omp.ubb99bpf |
| `session` | dev3.ubb99bpf |
| `agent_tool` | OMP |
| `tooling_profile` | dotfiles@931583a |
</details>

PR #1213, opened 2026-09-04
feat(buck2): transfer independent TypeScript library authority to Buck

schickling-assistant:
## Problem

Phase 3 TypeScript widening still dual-produces independent library packages in the root `tsconfig.check.json` / `tsconfig.emit.json` solutions. After notion-top (#1209) those solutions still listed 23 project refs. That keeps Buck and tsc as competing producers for packages that have no remaining intra-repo TypeScript dependents in the root solutions.

## Goal

Transfer typecheck and declaration authority for the remaining independent library packages to Buck. Each admitted package leaves both root TypeScript solutions. Public type conditions consume Buck declarations; runtime defaults stay at source.

## Decisions

Admitted in this layer: effect-path, kdl, oxc-config, npm-release, effect-ai-claude-cli, agent-session-ingest, effect-react, effect-rpc-tanstack, pty-effect, restate-effect, ci-tools, effect-schema-form, react-inspector.

Left out, with cause:

- **genie** — import cycle between the admissions registry and the generator.
- **buck2-tools** — the root `BUCK` still sources `package-tree.ts` / `typescript-runner.ts` from that package; a nested `BUCK` steals those files.
- **effect-rpc-tanstack/examples/basic** — route file `users.$id.tsx` fails the package-source safety alphabet.
- **context/effect/socket, context/opentui** — example trees, already `noEmit`; `:dist` tried to write `tsbuildinfo` into a read-only package tree.

Leftover project references on still-dual packages (kdl-effect, effect-schema-form-aria, megarepo) were deleted so oxlint type-aware checking against `tsconfig.check.json` does not pull admitted packages back into the solution.

The admissions unit test now compares authority entries sorted by package path. Registry order and workspace-package order are independent; the previous `toEqual` only passed while the two lists happened to agree.

## Verification

- `devenv tasks run genie:run` succeeded.
- `devenv tasks run lint:fix:oxlint` succeeded (0 warnings, 0 errors) after leftover project refs were removed.
- `devenv tasks run lint:fix:format` succeeded.
- `devenv tasks run genie:buck2:test` — 25 pass / 0 fail.
- `buck2:typescript:materialize-dist` reached BUILD SUCCEEDED (47 local actions) once.

`devenv tasks run check:quick` did not complete after that success: Watchman hit a non-recoverable inotify poison, then a retry spent ~25 minutes on `pnpm_importer` for notion-cli and was cancelled. This PR stays draft until `check:quick` is green.

## Complexity

[...]

PR #1209, opened 2026-09-03
feat(buck2): transfer notion top TypeScript authority to Buck

schickling:
Stacked on #1208. Buck becomes sole dist producer (noEmit x4) for notion-md/react/datasource-sync/cli; root solutions 27->23 refs.

Deletions: 16/16 inherited edges gone (12 top->base, 4 intra-base) + 15 source->dist sibling swaps. Known debt (documented, green): 4 intra-top refs left (react->md is test-only without a declared dep — needs resolution tracing before deletion), cli->effect-path (target not authoritative), dist-backed types conditions (dual resolution persists).

Deltas: root solutions -8 refs; -16 edges; +4 dist overlays; 8 dual producers eliminated. Proof: genie:check + genie:buck2:test (25/0); CI proves Buck typecheck targets. Evidence: context/buck2/02-execution/.experiments/2026-09-03-notion-top-authority-transfer.md. Interview: q6+q8.

Issue #1077, opened 2026-08-12
Make devenv shell activation independent of Buck and repository setup


schickling:
## Outcome

Make the steady-state development shell a mutation-free environment activation path. Buck warmth accelerates the first explicit repository command; it is never consulted during shell entry.

## Contract

```text
Nix/HM -> immutable machine capabilities
       -> mutation-free devenv shell
       -> explicit lazy Buck command -> stable daemon/DICE/cache
```

## Acceptance

- [x] shell entry runs no pnpm install, Genie generation, megarepo reconciliation, Buck command, or stage-0 realization
- [x] repo-local buck2-task uses source mode without rebuilding the distributable launcher
- [x] stage-0 tools resolve lazily, single-flight, with exact semantic invalidation and executable validation
- [x] warm shell and warm Buck are distinct benchmark workloads
- [ ] warm shell target: p50 <= 500ms, p95 <= 1s on an unloaded runner
- [x] warm Buck no-op target: p50 <= 100ms, p95 <= 250ms, zero actions
- [x] VRS and #1050 remain reconciled
- [x] implementation is appended to native stack #1074 as #1078

## Evidence

Implementation gates pass locally: `check:quick`, `genie:check`, setup module options, and 28 buck2-tools tests. The evaluated shell graph contains only devenv's own file tasks before `devenv:enterShell`; setup and custom OTEL shell tasks are absent. A real lazy stage-0 foundation build passed, followed by a warm no-op build. Warm resolver hits measured 170-200ms. Warm Buck status measured p50 11.4ms/p95 16.5ms; warm no-op foundation build measured p50 12.5ms with zero actions.

Full shell timing remains no-verdict: pinned devenv 2.2.1 probes did not exit reliably under host load averages 9-19. This is tracked as a devenv evaluation/cache acceptance gate, not attributed to Buck.

### 2026-09-08 whole-repository measurement

A steady-warm shell entry measured 18.4 CPU-seconds and 1.90 GB peak RSS, compared with 10.2 CPU-seconds and 1.05 GB on `main` (about 1.8x). Wall time was about 20 seconds per invocation. Across 22,597 measured tool turns at a 7.1-second median, repeated shell entry represented an estimated 44.6 aggregate hours. The warm-shell acceptance target remains open.

Parent: #1050
Cleanup coordination: #1065
Implementation: #1078
Buck authority follow-up: #1147
Current inert machinery: #1227

<details>
[...]

PR #1220, opened 2026-09-05
feat(buck2): complete whole-repository authority cutover

### LINKED ISSUE #1212: Buck TS action staging exceeds namespace runner disk/time budget as admissions grow (blocks #1209)
# Buck TS action staging exceeds namespace runner disk/time budget as admissions grow (blocks #1209)

Part of #1147. Companion: decision `0025` (CoW/reflink local disk economics) and `03-materialization/.delta/DELTA-001-assembler-hardlinks-pending-0025.md`.

## Motivation

Phase 3 widening is gated by CI lanes that can no longer run the admitted graph. PR #1209 (notion top authority: 4 packages, 16 edge deletions, root solutions 27→23) is red on `buck2` + `typecheck` while its content is proven good by every check that can run:

- stock `tsc -p` on all four packages: **5–8 s, clean** (md 8.3 s, react 5.5 s, ds-sync 7.5 s, cli 4.3 s)
- `genie:check` + `genie:buck2:test` green locally
- one retry of the failed lanes reproduced the identical failure deterministically — not flake, not luck

Each admission adds roughly **~1 GB retained output + ~1 min staging work per lane**, while the Buck lanes on namespace runners execute everything cold (`BUCK2_NO_REMOTE_CACHE=1`, off-tailnet by policy — see `genie/ci-workflow/shared.ts:57` and the "CI/off-tailnet" record in `05-composition/.experiments/2026-08-30-production-cp-a-member-e2e.md`), on small disks, under ~30–35 min lane timeouts. Without intervention every further admission makes this worse, and BUCK-R07 (budgets) / BUCK-R08 (disk anti-duplication) become unreachable. This issue exists so the capacity fix lands *before* the next admission, not after five more red cycles.

## RCA

The bottleneck is action **staging**, not compilation. Per admitted package, Buck runs:

1. `package_tree` (`packages/@overeng/buck2-tools/src/package-tree.ts`): `cloneTree` full-copies node_modules (`copyFileSync(..., COPYFILE_FICLONE)` — a full data copy wherever reflinks are unsupported, e.g. ext4/overlay), then `assertContainedSymlinks` walks the *entire* tree with one `realpathSync` per symlink.
2. `tsgo_emit` / `tsgo_typecheck` (`packages/@overeng/buck2-tools/src/typescript-runner.ts`, `runEmit`/`runTypecheck`): `cp` of the whole tree into a staging dir, then `makeTreeReadOnly` — a **sequential, single-threaded** `lstat`+`chmod` per file — then tsgo, then the same crawl in reverse at cleanup.
3. `buck-out` **retains every `package_tree` output**, so retained weight grows ~linearly with the admitted count.

Measured on dev3 (fast NVMe, lightly loaded) with a storybook-closure package tree (**927 MB / ~40k entries**):

| Step | Time |
|---|---|
| stock `tsc -p` (compile control) | **8 s** |
| full `emit` runner end-to-end (`typescript-runner.ts` + pinned tsgo) | **41 s wall** (≈6 s compile, **≈35 s staging**: cp + chmod crawl + cleanup) |
| `package_tree` symlink-audit walk alone | **35 s** |

CI evidence (all on the same action family, both lanes):
- run `33806232240` → `ENOSPC: no space left on device` staging a node_modules tree ([typecheck job](https://github.com/overengineeringstudio/effect-utils/actions/runs/33806232240/job/100817247490)), with `Commands: 279 (cached: 0, remote: 0, local: 279)`.
- run `33809913585` → `notion-md:dist` (`tsgo_emit`) stuck 12+ min, lane canceled at ~33 min ([typecheck](https://github.com/overengineeringstudio/effect-utils/actions/runs/33809913585/job/100829079887), [buck2](https://github.com/overengineeringstudio/effect-utils/actions/runs/33809913585/job/100829080028)).
- rerun of both failed lanes → identical stall on the identical action ([typecheck](https://github.com/overengineeringstudio/effect-utils/actions/runs/33809913585/job/100938826451), [buck2](https://github.com/overengineeringstudio/effect-utils/actions/runs/33809913585/job/100938826675)).

Excluded with evidence: compile pathology (8 s control), md-specific content (tui-react carries the same storybook closure and builds fine — md:dist was simply the action running when the runner fell over), source/config drift (`genie:check` + unit tests green).

Adjacent hazard found during repro (not the CI bug, recording it here): pnpm's `node_modules` contains symlink cycles, so any *dereferencing* copy explodes — a `cp -rL` trial hit **71 GB** in minutes before being killed. Both runners correctly use `dereference: false`; this is a loaded gun next to the code, not the fired one.

## Local repro (dev3-class machine)

```bash
# 1. compile control — expect ~8 s, clean
node node_modules/.pnpm/typescript@6.0.3/node_modules/typescript/bin/tsc \
  -p packages/@overeng/notion-md/tsconfig.json --pretty false

# 2. package-tree audit walk — expect ~35 s on fast disk (scales with disk pressure)
time bun packages/@overeng/buck2-tools/src/package-tree.ts \
  --output /tmp/pt-repro/out --node-modules node_modules \
  --file package.json package.json

# 3. full emit action — expect ~40 s wall, ~6 s of it compile:
#    assemble <tree>/ with package files + node_modules + .workspace-siblings,
#    then:
time bun packages/@overeng/buck2-tools/src/typescript-runner.ts emit \
  /nix/store/2dnkwcaq8cysibwlfvh2zjwabl67wsa2-effect-tsgo/bin/tsgo \
  <tree> tsconfig.json dist src/mod.d.ts <out>
```

(Use hardlinked input assembly, `cp -al`, so input prep itself stays instant. Clean up `/tmp/pt-repro`, `/tmp/emit-repro` afterwards — the experiment is disk-hungry by nature.)

## Options and tradeoffs

Baseline: ~35 s staging overhead per action on fast local disk; ~1 GB retained per tree; 279 cold-local actions in the current graph; namespace lanes pay a large pressure multiplier and die at ~30–35 min / small disk.

| Option | Mechanism | Expected effect | Cost / risk |
|---|---|---|---|
| **A. Hardlink/reflink staging** (lands 0025 + DELTA-001) | share tree bytes via CAS/hardlinks instead of per-action full copies | copy + chmod crawl → ~0; retained disk collapses to unique bytes; helps every future admission *and* dev machines | medium: isolation semantics need design (chmod mutates shared inodes — read-only bind mounts / overlayfs / CoW layers); touches `buck2-tools` runners + `materialization.bzl` |
| **B. Self-hosted Buck lanes** | run `buck2`/`typecheck` where disk + warm cache live, per BUCK-R06/R07 end-state (cf. #1056 milestone: 283/283 cached, 30.6 s fresh-context) | unblocks the whole roadmap immediately; lane times collapse with cache reads | runner topology + tailnet; dev3 root pressure (#2293 — 81% full and climbing under concurrent agent load at time of writing) must be handled first or ENOSPC just moves address |
| **C. Raise lane timeouts / bigger namespace runners** | buy headroom | exactly one round of relief; does not change the per-admission curve | recurring cost; re-litigated every few admissions |
| **D. Retain less** (buck-out GC between phases, fewer materialized intermediates) | cap peak disk | saves disk, not time; partial relief only | risks invalidating debugging/caching assumptions; needs measurement per lane |

Recommendation: **B unblocks now, A fixes the curve**; C only as a stopgap. Whichever lands, #1209 (open, content-proven, red purely on capacity) is the validation vehicle: it goes green on the chosen path with no content changes.

## Acceptance

- [ ] #1209 green without content changes on the chosen path
- [ ] per-admission staging budget recorded (p95 action staging time, peak lane buck-out) and added to the R07/R16 benchmark record
- [ ] decision record or DELTA entry for the chosen option (amend 0025 or file the successor)

schickling-assistant:
## Problem

The repository had two build authorities. pnpm owned installation and TypeScript
project graph; Nix owned eight `pnpm-deps` fixed-output producers that each
copied a per-consumer closure. Nothing owned the boundary between them, so the
same dependency set was materialized repeatedly, product bytes were not
platform-invariant, and no sandbox contained an action. There was no single
answer to "what does this project depend on, and what exactly did it produce".

Four failures showed the same missing-owner shape once the cutover ran on CI:

- the generated `.buckconfig` pins `buck2.file_watcher = watchman`, but nothing
  guaranteed a `watchman` binary existed; on a runner (and any bare-shell caller)
  PATH has none, so every Buck invocation failed to start its file watcher
- Bubblewrap refuses `pivot_root` inside a default unprivileged Namespace
  container, so every sandboxed Buck TypeScript action died with
  `bwrap: pivot_root: Operation not permitted` (observed on `lint:check` in run
  34047874617)
- the cold-bootstrap proof runs the Buck-built Genie product, whose working
  directory is the composed workspace root while the tree under generation is a
  separate install-free `git archive` export; every projector that read
  repository-relative paths through `process.cwd()` read the wrong tree
- that same cold check ran the whole generator set, including `design-time`
  generators that use the installed runtime graph by declaration (decision 0004),
  so importing one from a tree with no `node_modules` resolved an ambient
  dependency instead of proving anything

## Goal

One authority. Buck2 owns the whole repository graph: 38 TypeScript projects,
their tests, tools, editor views, and package consumers; Nix imports strict,
platform-invariant products from it and never re-derives them. Every action runs
inside a sandbox (Bubblewrap on Linux, Seatbelt on Darwin). A cold, install-free
tree can regenerate its own bootstrap projections and be proven byte-identical.

## Decisions

The first four repairs on top of the cutover, each fixing one of the failures
above:

| commit | fix |
| --- | --- |
| `a2c307a45` `fix(buck2): provision Watchman from the composition root` | the generator that pins `file_watcher` now provisions and proves the binary, so Buck starts its watcher on a bare shell |
| `77d38bacd` `fix(ci): enable privileged runners for Buck sandboxes` | Buck-bearing jobs request `container.privileged=true` in the Namespace syntax their label kind requires, so Bubblewrap can `pivot_root` |
| `f15d009f5` `fix(ci): make cold bootstrap independent of process cwd` | projectors anchor on their owning repo instead of `process.cwd()`, so the cold export is the tree actually generated |
| `60a1d4a31` `fix(ci): scope cold proof to bootstrap generators` | the cold check runs only `bootstrap`-phase generators, with a non-vacuity assertion so the narrower scope stays a real proof |

Canary run #5367 got past those and exposed a second, distinct set of
regressions — none of them the Watchman or Bubblewrap failure. Grouped by cause:

| group | commits | fix |
| --- | --- | --- |
| Buck query lanes | `489f29707`, `bc427981f` | `buck2 build`/`test`/`run` parse every positional as a target pattern, never a query expression, so both query lanes died with `Parsing target pattern filter(...)` / exit 3 (run 34067718982). A generator helper expands the query through `uquery`, refuses a vacuous expansion, then builds the resulting labels; bundle-smoke switches `test` → `build` and folds its hand-listed literal into the query, since `package_bin_check` returns no `ExternalRunnerTestInfo`. Separately, `mapfile -t x < <(cmd)` reported `mapfile`'s status, making a *failed* query indistinguishable from an empty one — capture now propagates the producer's exit and the emptiness check runs only on successful output. |
| Cache policy | `af187f56f` | `ExternalRunnerTestInfo` carries its own executor, and `_test_info` wired all three cache switches to the `cacheable` attr, so under `BUCK2_NO_REMOTE_CACHE=1` every deterministic lane still requested remote lookup and died with `No engine address`. Participation is now determinism **and** root policy, read through one exported pair of root-config readers; determinism refusals stay on the raw attr so root policy can never make a non-reproducible lane cacheable. |
| Read-only trees | `6228d76ab`, `1408c2c53` | Buck publishes read-only trees, and both consumers assumed writable inodes: `git worktree remove` failed `Permission denied` and stranded the job-local store (job 101579327940), and hardlinking a read-only override source handed output finalization an inode it cannot chmod (node-pty `EPERM`). Cleanup now `chmod -R u+w` *after* every ownership guard and refuses to fall through on failure; overrides are copied and given owner write on top of their declared mode bits, never mutating the declared source. |
| Cold-tree ordering | `a1310319b` | Type-aware oxlint resolves types through the Buck-published editor views, which a warm tree hides. The dependency edge sits on the oxlint task itself, not the `lint:check` aggregate, so a direct task run is ordered too. |
| Test hermeticity | `a8712dfd0` | The ts-diagnostics fixture stubbed `otel-span` on PATH only while wrappers resolve bridges via `OTEL_SPAN_BIN`/`OTEL_SCRAPE_BIN`, so an ambient scrape could wrap the real compiler and an ambient endpoint could be dialed. Both bridges are pinned to the test's stubs, endpoints cleared, and failures now print captured output instead of aborting with the stream trapped. No assertion weakened. |
| Shell-entry cost | `c39e27426` | `weaver:live-check` took the weaver binary and semconv model as already-realized derivations, pulling the Rust build and the upstream FOD into `devenv-shell-env`: every unrelated lane paid for them on shell entry. Both are now resolved at task time from a path flake ref, matching `weaver:check`/`weaver:diff`. Unavailability blocks rather than degrades — a skipped e2e would report success without validating the emitted OTLP. |
| Lint / comment truth | `34f6acce1` | `explicit-boolean-compare` on the cleanup fixture's teardown guard, and the Weaver comments still promised a block-vs-degrade split the task no longer has. |

Canary run #5368 went further still — `typecheck`, `test-megarepo-cold-gc` and
the Buck lanes passed — and the next six repair groups come from what its
`bundle-smoke`, lint and macOS legs exposed:

| group | commits | fix |
| --- | --- | --- |
| Package-command cwd/argv | `5ac6213b9` | A local Buck action receives project-relative artifact paths and runs from the project root, but the package command runner launches its child with the package tree as cwd, so the entrypoint, `{OUT}` and `{TREE}` were resolved a *second* time against the tree and every `package_bin_check` died with `Module not found <tree>/<tree>/<entrypoint>`. `planPackageLaunch` resolves tree, output and placeholders into one pure launch plan before the directory change; `resolve` is a no-op for the absolute paths `buck2 run` hands an exec launch, so both callers share one contract and package-relative module/config resolution keeps its cwd. |
| Source-smoke runtime | `ee97f075e` | The notion-db source entrypoint imports `node:sqlite`, which Bun cannot resolve, so its bundle smoke failed although the published product already launches Node 24. A product's bundle target and the raw source a smoke executes are different artifacts, so the smoke runtime is declared separately: `JavaScriptProduct` gains `smokeRuntime` (default `bun`), every generated `package_bin_check` renders it explicitly, and the rule resolves the admitted host node capability for the child while the runner keeps executing under Bun. |
| Undeclared capabilities | `b9e1ad49a`, `5b891abb5` | The hermetic Vercel e2e stub ran bare `mkdir`, `tr` and `cat`; containment exposes only declared capability executables, so three build-mode cases died with exit 127 `mkdir: command not found`, and the alias-retry counter's `cat … \|\| echo 0` would have silently pinned attempts at one — a retry assertion passing on hollow evidence. All three are now declared `gnu/coreutils/v9` capabilities threaded into the ci-tools test target and called by attested absolute path. |
| Unrealized fixture closure | `49794e766` | The diagnostics-parser test extracted the shipped `ts:check` exec with `nix eval --raw`, which prints embedded store paths as inert text and drops Nix string context, so nothing realized them; on a cold machine the script hit an unrealized `bc` and died on its first arithmetic. The exec is built through `pkgs.writeText` so string context makes every embedded tool an input derivation, and a realization sentinel names a still-missing tool instead of surfacing `bc: not found` from inside the parser. |
| macOS host assumptions | `d6733d85f`, `d1f880651` | The JavaScript runner reached `sandboxInvocation` without `darwinMetadataLinks`, so every Seatbelt lane failed closed before a test could execute and an unadmitted Darwin kernel was never rejected; it now mirrors the TypeScript runner, requiring an admitted Darwin release and resolving the canonical OS metadata spellings. Separately BSD `chmod` has no `--` marker: it read the marker as a file operand, exited non-zero, and aborted composition cleanup under `set -e`, stranding the job-local store. The marker is dropped (the store root is an absolute resolved path that can never parse as an option) and covered by a fixture that puts a BSD-faithful `chmod` first on PATH. |
| Silent lint attribution | `630d9333f`, `2a6658768` | A nonzero oxlint/oxfmt status could arrive with no stdout and no stderr at all: `xargs` collapses child status 1..125 to 123, and a setup/scan step failing under errexit aborted the exec anonymously. `mkLintExec` now carries lane identity, a stage marker and an ERR trap, counts streamed stdout bytes, and reports lane/stage/status/file-count/stdout-bytes/command on one stderr line before re-raising the *same* status — `stdout_bytes=0` names a tool that died mute. The first cut of that accounting named the inherited descriptor as a path (`tee /dev/fd/3`), which on Linux *reopens* the file with `O_TRUNC` and destroyed prior bytes of any redirected CI log; the real stdout is now reached with `>&3` and the counter reads a FIFO with `wc` waited on before the count is read. Findings still stream verbatim; `denyWarnings` and type-aware behavior are untouched. |

- **Normalized store + assemblies over per-consumer closures.** pnpm closure
  copies are replaced by normalized store entries, five SCC assemblies,
  metadata-only importer views, and real platform selection. Per-consumer copies
  were the direct cause of the duplicated fixed-output producers.
- **Watchman is provisioned by the generator that pins it.** The generator that
  owns `file_watcher` now also owns the binary: `MR_COMPOSITION_WATCHMAN_BIN`
  pins the exact Nix realization, the composition runtime proves it is a real
  executable *file* (not merely `X_OK`, which also passes for a searchable
  directory), and the generated wrapper prepends its bin directory without
  dropping any caller PATH entry. Putting Watchman on the ambient PATH instead
  would have made the config's correctness depend on the caller's environment.
- **Privileged-container syntax is derived, not hand-written.** Namespace spells
  the feature in exactly one of two non-interchangeable places depending on label
  kind: a `namespace-profile-*` label carries it as a `;`-suffix of the profile,
  a raw machine label must stay verbatim and carry it as an entry of the
  companion `namespace-features:` label. The workflow generator derives the right
  form and *refuses* privileged mode on a raw label that cannot honor
  `namespace-features`, rather than emitting a pair Namespace silently ignores.
  The matrix value stays the bare profile, so the required check context
  `test (namespace-profile-linux-x86-64)` is unchanged; only the resolved
  `runs-on` carries the feature, and only on the Linux leg. Jobs that run no Buck
  action keep the default unprivileged container.
- **Projectors anchor on their own repo, not on `process.cwd()`.** Each projector
  resolves through `defineRepoContext`, and `modulePathFromUrl` recovers a
  module's repo-relative identity when the compiled product stages its import
  graph into a temporary mirror. Repo-root resolution accepts a megarepo config
  *before* `.git`, because a cold export legitimately carries no `.git`; the
  root-test staged tree therefore also stages `megarepo.kdl`, otherwise the copy
  under `buck-out` resolves out to the live composition root and reads bytes the
  action never declared. The cold proof also stopped resolving its TypeScript
  compiler from PATH and now reads the exact pinned realization from the member's
  own `.buck2/capabilities` projection — the same `effect-tsgo` Buck's own
  TypeScript actions execute — failing closed on an absent projection, a
  non-store path, or a non-executable target.
- **Cold proof is scoped to `--phase bootstrap`, with a non-vacuity assertion.**
  Only bootstrap generators can honestly run against an install-free tree, so the
  two pnpm lock projections declare `// @genie-bootstrap` and the check runs that
  phase alone. An assertion over the check's JSON events refuses a pass that never
  reached `buck2/dependencies/BUCK` or `buck2/dependencies/pnpm-lock.sha256.json`,
  so the narrower phase cannot decay into an empty proof.
- **Cache and capacity lanes are opt-in.** Shared cache and capacity evidence run
  on Namespace as separate lanes plus recurring multi-platform sandbox gates, so
  the default PR path never depends on a cache being warm. The cache is
  evictable; Git remains the durable authority for every product byte.

### Completed locally — head `5b891abb5` unless noted

| proof | result |
| --- | --- |
| `CI=1 devenv tasks run check:quick --no-tui` (final run, after the #5368 repairs) | pass, 16m48 |
| exact 11-target bundle-smoke query expansion, built locally | pass, all 11 labels green |
| `CI=1 devenv tasks run check:quick --no-tui` (at `34f6acce1`) | pass, 11m37 |
| `CI=1 devenv tasks run check:quick --no-tui` (at `60a1d4a31`) | pass, 23m33 |
| `bootstrap:cold-proof` (committed; install-free `git archive` tree, at `60a1d4a31`) | pass, 14m32 |
| `buck2 test effect_utils//packages/@overeng/megarepo:test` | pass, 60 test files |
| `buck2 test effect_utils//packages/@overeng/genie:test` | pass, 29 test files |
| `buck2 test effect_utils//genie/buck2:test` | pass, 16 test files |
| `CI=1 devenv tasks run test:devenv-modules --no-tui` | 60 ok |
| cold proof via the direct composition wrapper (bare PATH, no devenv) | pass — Watchman resolved from the pinned realization |
| strict product publication + exact `reconcile.sh check` | pass |
| strict JavaScript importer negative/positive contract suite | pass |
| dependent `schickling/dotfiles` cache-auth IaC | locally converged, 0 drift |

Before → after on the two CI-observable failures:

```
watchman   before  Buck fails to start its file watcher (no watchman on PATH)
           after   wrapper prepends the pinned realization; cold proof runs on a bare shell

bubblewrap before  bwrap: pivot_root: Operation not permitted   (run 34047874617, lint:check)
           after   Buck-bearing jobs request container.privileged=true via the derived label form
```

### Remote proof — canary #5368 failed, #5369 in progress

**#5367** at `60a1d4a31` did not pass. It cleared the Watchman and Bubblewrap
failures it was meant to prove and then failed on a second, unrelated set of
regressions the warm local tree had hidden: both Buck query lanes rejected their
query expression as a target pattern (exit 3, run 34067718982); every
deterministic test lane requested remote cache under `BUCK2_NO_REMOTE_CACHE=1`
and died with `No engine address`; job 101579327940 stranded its job-local store
because `git worktree remove` cannot unlink a read-only published tree; and a
read-only override source broke node-pty entry assembly with `EPERM`. Cold-tree
lint ordering, a non-hermetic diagnostics fixture, and a `devenv-shell-env`
closure carrying the Weaver build surfaced alongside.

**#5368** at `34f6acce1` also did not pass, but it was the most evidence-rich run
so far: `typecheck`, `test-megarepo-cold-gc` and the Buck lanes passed, and the
failures it did produce were each a real, distinct defect rather than noise —
package-command cwd/argv double-resolution, the notion-db source-smoke runtime,
undeclared `cat`/`mkdir`/`tr` in the ci-tools fixture, an unrealized ts-fixture
closure, macOS Seatbelt OS-metadata and Darwin release admission, BSD `chmod`,
and lint failures that arrived with no attribution at all. All are repaired in
the Decisions tables above.

The bounded canary is now the single automatic PR workflow run **#5369** at head
`5b891abb5`, gating manual fan-out on the same four repaired-path jobs. **Run
#5369 is in progress; nothing below has passed yet.**

- [ ] canary run #5369 green on all four jobs: `typecheck`, Linux `test (namespace-profile-linux-x86-64)`, `bundle-smoke`, `test-megarepo-cold-gc`
- [ ] DQ1 — Namespace Linux x86_64 shared-cache upload / restore / outage proof
- [ ] DQ4 — Namespace Linux x86_64 capacity pair and numeric envelope
- [ ] `devenv tasks run check:all`
- [ ] evidence join across the lanes
- [ ] Linux ARM64 and Darwin ARM64 sandbox/product byte-identity proof
- [ ] post-merge: three-host BRIDGE-R12 convergence
- [ ] post-merge: dependent `schickling/dotfiles` lock repin to the merged SHA

## Complexity

This PR is large because the cutover cannot be partial: two authorities cannot
both own the project graph while a migration is in flight, so the root install
and both root TypeScript producers are deleted in the same change that makes
Buck able to replace them. Within that, the added mechanisms each buy something
that cannot be bought more cheaply:

- deriving the privileged label form is more code than hard-coding it, but the
  two spellings are not interchangeable and a wrong pair fails silently
- `defineRepoContext` / `modulePathFromUrl` exist because a compiled product's
  cwd is structurally not the tree it generates; no ambient convention can fix that
- the non-vacuity assertion exists because narrowing a check's scope is exactly
  how a check quietly becomes a no-op

No new runtime dependency is introduced for consumers.

## Concerns

- Privileged containers are a real privilege increase on the Linux Buck-bearing
  jobs. It is scoped to jobs that execute a Buck action and refused where the
  label cannot express it, but it is still a wider blast radius than the default
  unprivileged container.
- Cache lanes are opt-in and the cache is evictable. Correctness must never
  depend on a warm cache; Git is the durable authority. A cache outage should
  degrade to slow, not to wrong — that is exactly what DQ1's outage leg checks,
  and it has not run yet.
- The cold proof is now narrower by construction (`bootstrap` phase only).
  `design-time` generators are, by decision 0004, not provable from an
  install-free tree; the non-vacuity assertion bounds the risk but does not
  remove it.
- Non-Linux sandbox parity (Linux ARM64, Darwin ARM64) is still unproven on this
  head.

## Friction & bottlenecks

- *Friction* — Namespace's privileged-container feature has two spellings that
  depend on label kind, and the wrong one is accepted and ignored rather than
  rejected. Diagnosing this cost a full CI round-trip; the generator now encodes
  the rule so it cannot be re-learned.
- *Friction* — `bwrap: pivot_root: Operation not permitted` surfaced only as a
  failing sandboxed action deep inside `lint:check`, with no signal that the
  container, not the action, was at fault.
- *Friction* — a warm local tree hides a whole class of failure: published
  editor views already on disk, writable inodes left over from an earlier run,
  a remote cache never consulted, and every tool the sandbox does not declare
  still on PATH. Nearly every regression canaries #5367 and #5368 found was
  invisible to a green local `check:quick`, which is the argument for the cold
  proof and for keeping the canary bounded rather than trusting local green.
- *Friction* — `mapfile -t x < <(cmd)` reports `mapfile`'s status, so a failing
  Buck query was indistinguishable from an empty graph and the real diagnostic
  was discarded. Cheap to write, expensive to debug.
- *Friction* — lint lanes could fail with no stdout, no stderr and a status
  `xargs` had already collapsed to 123, so a red run named nothing at all.
  Attribution had to be built (`630d9333f`) before the actual findings could be
  read.
- *Friction* — naming an inherited descriptor as a path (`tee /dev/fd/3`)
  silently reopens the underlying file with `O_TRUNC` on Linux, destroying every
  byte a redirected CI log had already collected. Found only by planting a
  sentinel before the stage (`2a6658768`).
- *Friction* — `nix eval --raw` prints embedded store paths as inert text and
  drops string context, so a fixture that looks fully pinned realizes nothing
  and fails on a cold machine with a bare `bc: not found` (`49794e766`).
- *Friction* — macOS-only breakage (BSD `chmod` rejecting `--`, Seatbelt OS
  metadata) is only reachable from a Darwin runner, so each iteration costs a
  full remote round-trip.
- *Bottleneck* — the local proof loop is long: `check:quick` ran 23m33 at
  `60a1d4a31`, 11m37 at `34f6acce1` and 16m48 at `5b891abb5`, with
  `bootstrap:cold-proof` a further 14m32, so a repair costs tens of minutes of
  local wall time before a remote round-trip even starts. That is why the remote
  step is a bounded four-job canary rather than a full fan-out.
- *Bottleneck* — `weaver:live-check` pulled the Weaver Rust build and the
  upstream semconv FOD into the closure of `devenv-shell-env`, so both were
  realized on every shell entry before any task ran and every unrelated lane
  paid for them (`c39e27426`).

## Follow-ups

- Cross-platform sandbox/product byte-identity for Linux ARM64 and Darwin ARM64,
  covered by the recurring multi-platform sandbox gates rather than the PR path.
- Post-merge three-host BRIDGE-R12 convergence and the dependent
  `schickling/dotfiles` exact-SHA lock repin — both require a merged SHA and are
  out of scope for this PR.

## References

Closes #1212

Supersedes #1209 and #1213.

Dependent cache-auth IaC: `schickling/dotfiles#2546`.

<details>

| field | value |
| --- | --- |
| `agent_identity` | dev3.direct.omp.k7stwr6d |
| `session` | dev3.k7stwr6d |
| `agent_tool` | OMP |
| `tooling_profile` | dotfiles@7534055 |
</details>


schickling-assistant:
## Closing report — PR #1220 review and re-cut decision (2026-09-08)

This PR is superseded by a two-PR stack (tracked in #1147 → "Active delivery"). It stays open as
the reference until PR-A is ready. This comment records what the review found and what both the
authoring seat and the review seat learned, so the next attempt does not re-derive it.

### What the review measured

Evidence was produced by independent sub-agents (transcript study, diff anatomy, VRS history,
external precedent research, adversarial complexity review, main-vs-branch benchmarks in isolated
worktrees, and a wall-clock/bottleneck accounting of the authoring session). Highlights:

| finding | number |
|---|---|
| Diff | 1,370 files, +1,132,692 / −54,798; **93.6 % generated payload** (41 MB product bytes in `nix/buck2-products`; an 873-file `.editor-view` tree tracked *and* gitignored); authored surface ≈ 34k lines, ≈ 75 new named concepts |
| Buck2 engine vs vision criterion 3 | whole 38-project warm no-op **0.08–0.57 s**; one leaf edit → 11/38 projects in **4.38 s** — the engine meets the budget |
| Shipped loop | `check:quick` **9.5–49 min**, red 6 of 9 locally; `buck2:check` **failed 9/9 from a fresh composed checkout** (stale tracked products; untracked products break Nix eval; `.editor-view` references a `.backing` layer git does not carry). `main`: 66 s / 57 s no-op, green |
| Where the time went (94.5 h) | sub-agent fan-out waits 59 %; waiting on human answers 21 % (7 of the 10 longest operations); local gates 9 %; CI waits 1 % |
| Structural sinks | serial editor-view publication (one run = 80 % of a 61-min gate); products rebuilt `--no-remote-cache` on every `check:quick`; per-job cold CI composition = 54 % of canary minutes; `devenv shell` entry ×1.8; whole-repo tsgolint OOM on 16 GiB runners; `Cache hits: 0%` whenever Buck ran outside the devenv shell |
| VRS gates for this flip | DQ1, DQ4 `blocked` at head; BUCK-R15 ledger measured once (+32,290 / −973, 2026-09-01), never for the largest phase; criterion 3 never measured on the whole surface |
| Precedent | upstream Buck2 does not sandbox local actions (README; facebook/buck2#358); the largest public TS/pnpm-on-Buck2 ruleset is ~1.5k lines with pnpm as install authority and nothing committed |

What is genuinely good and is preserved: the normalized dependency store + SCC assemblies
(decisions 0022/0030: per-consumer layer 7.38 MB vs 5.72 GiB; 416/416 cache hits incl.
relocation), the Nix product-import contract, the cache-policy readers, the uquery lane helpers.

### Decisions (Johannes, 2026-09-08, review seat `dev3.direct.omp.5bnx6kwm` q1–q6)

1. Whole-repository Buck2 authority stays; DQ1/DQ4/BUCK-R07/BUCK-R15 become merge gates of the flip.
2. Products ship as immutable release artifacts + a small tracked digest manifest; no product bytes in git (amends 0031).
3. No per-action bwrap/Seatbelt sandbox on the PR path; containment is a periodic/transfer-time gate; zero privileged PR jobs (amends BUCK-R04).
4. Land as PR-A (inert machinery) + PR-B (small atomic flip), stacked (amends 0030-A1).
5. Authoring seat parked after push + handoff; a fresh seat re-cuts.
6. PR-A also fixes the two sinks that stand between the engine's 0.5 s and a passing gate (parallel per-package editor-view actions; cache reachable outside the devenv shell); the CI-composition, shell-entry (#1077) and tsgolint-sharding sinks are filed as measured issues.

### Lessons — review seat

- **Measure the gate the user runs, not the engine.** Every recorded budget number was a subset
  (8 projects, 10 products); the first whole-surface number was this review's. A budget that is
  never measured on the real surface is a budget that is asserted.
- **Green-locally is not evidence when local state is unreproducible.** 33 GB of seat-local
  `buck-out`/`.editor-view` made the branch green only on one machine. A fresh-clone gate is the
  cheapest defence and must exist before any authority flip.
- **The diff size was a symptom, not the disease.** 94 % of the lines were generated; the real
  cost was ~34k authored lines, 75 concepts, and a 20-layer path from "edit a file" to "CI green"
  (3–4 on main). Judge complexity by layers a contributor must understand, not by `+/-`.
- **Scope grew through agent-authored binary menus.** Four forks in 30 h (full store redesign → whole
  repo → tracked products → strict portable product) were each posed with the larger option framed
  as principled and the "stop here / ship the narrow fix" option never re-offered, although it was
  the agent's own first recommendation. Rule: every decision menu carries a smallest option, and
  the recommendation is stated first.
- **Challenges after the fact do not move architecture.** Three explicit simplicity challenges
  produced three evidence rounds and one decision record ratifying the existing choice. When a
  design challenge arrives after the payload is committed, the right move is a re-cut proposal,
  not a defence.
- **Patch-forward has a signature.** 33 of 45 commits were same-day repairs; five canaries with a
  monotone failing set; 114 of 285 sub-agent yields were Fix/Diagnose/Recover. Two failures of the
  same class on the same surface should stop the loop and force a root-cause write-up.
- **Fan-out is not throughput.** 178 sub-agent spawns; 59 % of wall clock spent waiting on them;
  more `hub` calls than file reads+edits+writes. The aggregate per-turn latency (22k turns × 7 s)
  exceeded all build time combined — host contention (dotfiles #2377) inflated everything 2–4×.
- **Human answer latency is a first-class cost.** 21 % of the window waited on `ask`; batch
  questions, proceed on refinements with a recorded assumption, and never park a seat on an
  operational question for six hours.
- **The requirement that encodes "simplify" must be run, not written.** BUCK-R15 was designed to be
  the running total and was executed once. Make the ledger and the budgets a CI-rendered table on
  the PR, so drift is visible per push.

### Lessons — authoring seat (`dev3.direct.omp.k7stwr6d`)

- Start with a genuinely clean checkout and make it the first executable acceptance test. My warm store hid ownership and bootstrap defects until late.
- Define the artifact boundary before optimizing publication. Source, tracked product, Nix input, and machine-local editor view must each have one owner and one transition.
- Never commit machine-local editor-view stores. A Git checkout cannot preserve their read-only invariant or their complete backing graph.
- Add Watchman OS-level ignores when the composition chooses Watchman, not after scale makes recrawls visible. Watch scope is architecture, not tuning.
- Drive Buck-native `:typecheck` early for Effect code. Vanilla `tsc` missed TS3770xx Effect diagnostics that the real Buck `effect-tsgo` gate rejected.
- Stabilize source before republishing products. Repeated source/product iterations created unnecessary bundles, commits, reviews, and history rewrites.
- Use one clean, full-fidelity vertical slice early: fresh checkout -> Nix eval -> product check -> editor publication -> `check:quick`. Targeted tests are necessary but did not prove the composed system.
- Preflight shared runners and resource pressure before long gates. Do not overlap equivalent heavy work, and do not infer pressure from swap occupancy alone.
- Preserve failure evidence, but stop after the first representative failure. Several long editor-publication runs repeated a problem that a clean-checkout model would have exposed faster.

Ranked blockers it left for a green `check:quick` at `4c28e6e0d` (its words, abridged): (1) tracked
machine-local editor-view snapshots that a fresh checkout cannot satisfy; (2) tracked products must
match the source and untracked bundles are invisible to Nix evaluation; (3) bootstrap order must not
depend on a warm `buck-out`/Nix store/`.editor-view`. Its Watchman fix (deterministic root
`.watchmanconfig` with derived `ignore_dirs`: initial crawl 287 s → ~130 ms) and partitioned lint
(39 units, 1.85 GiB peak, no OOM) are valid inputs for PR-A.

The branch's final heads are preserved as `backup/k7stwr6d-final-4c28e6e0d` and
`backup/k7stwr6d-post-cancel-c401b4eaa` (this branch's remote head `d001b22c7` predates yesterday's
history rewrite; a non-force push was correctly refused).


### Pointers

Evidence files are seat-local review artifacts (paths in #1147); the durable record is this
comment, #1147, the VRS amendments in `context/buck2`, and the decision records in the review
seat's tree.


<details>

| field | value |
| --- | --- |
| `agent_identity` | dev3.direct.omp.5bnx6kwm |
| `session` | dev3.5bnx6kwm |
| `agent_tool` | OMP |
| `tooling_profile` | dotfiles@703d931 |
</details>

schickling:
PR-A is now ready: #1227. It carries the inert Buck2 machinery, ten immutable v3 product releases, the manifest-pinned Nix boundary, complete release-attestation/Nix realization proof, and green fresh composed CI run 34388447244. The authority flip remains isolated to PR-B after #1227 merges. #1209 remains untouched.

PR #1331, opened 2026-09-21
fix(buck2): drop sibling wiring from from-source package trees

schickling-assistant:
## Problem

The first `publish-products` run on `main` after #1321 (run 35628260764) failed while realizing `.#buck-product-genie-from-source`:

    Error evaluating build file: `effect_utils//packages/@overeng/genie:BUCK`
        error: Argument `workspace_dependency_views` occurs more than once
           --> buck2/materialization.bzl:150:5

`nix/buck2-products/from-source.nix` rewrites the product's `package_view(` into `package_tree(` and only stripped the *empty* `workspace_dist = {}` / `workspace_dependency_views = {}` forms. Since projection v11 (#1283) 29 packages carry non-empty sibling maps, which fall through `**kwargs` into `_package_tree`'s hard-coded `workspace_dependency_views = {}`. The P1 lane proved the recipe on oxc-config only, which has no workspace siblings.

## Goal

Buck evaluation of every from-source product recipe succeeds for packages with workspace siblings.

## Decisions

Strip both blocks whole instead of teaching `package_tree` to accept them: in the sandbox the prepared tree already contains every workspace sibling as a source copy with its projected `node_modules`, and package exports resolve to `src/` at runtime (`"default": "./src/isomorphic/mod.ts"`), so the composed-workspace sibling wiring has no counterpart there.

## Verification

Regex check against the real genie BUCK block (129 → 113 lines, no `workspace_` left, other attributes intact):

    before: 129 after: 113
    name / dependency_view / files / runtime / runtime_entry / visibility
    still has workspace_: false

`nix build .#buck-product-genie-from-source` on this branch: Buck evaluation passes (before: the error above) and the build proceeds to the bundle action. That action then fails for a second, independent reason — see Concerns.

## Complexity

No new complexity.

## Concerns

This fix is necessary but not sufficient for `publish-products`. The bundle step fails next with `Could not resolve: "effect", "@effect/platform-node"`: the recipe's `preparedDeps` is the gh-ci-utils pnpm FOD, scoped (#1162) to gh-ci-utils' own workspace closure, so 13 of the 21 product packages have no install root in it. That is a design choice (dependency root scope) tracked as decision Q57; it is deliberately not bundled into this PR.

## Friction & bottlenecks

- GitHub Releases downloads intermittently 504 from the CI runners today; the retry helper did not classify `cannot download … from any mirror` as transient (fixed in #1329).
- `gh run view --log-failed` is broken on the operator host (root-owned gh-ci-utils config); the raw jobs API was used instead.

## Follow-ups

- Q57: dependency root for the from-source recipes (products-scoped prepared root vs. Buck dependency views from a Nix-realized store).

## References

- Refs #1321, #1283, decision 0037.

<details>

| field | value |
| --- | --- |
| `agent_identity` | dev3.direct.omp.t3mkm4sd |
| `session` | unknown |
| `agent_tool` | OMP |
| `agent_tool_version` | 18.1.14 |
| `agent_runtime` | OMP 18.1.14 |
| `tooling_profile` | dotfiles@5d1e05f |
</details>


schickling-assistant:
Superseded by decision q62 (2026-09-22): the from-source adapter is deleted in favour of the declared Buck graph fed by per-digest archive inputs (see explore.nix-store-views / explore.global-max-acquisition reports). The five-line sibling fix disappears with the adapter.

